PR #7, opened 2015-11-10
Add Travis CI badge [ci skip]

kuraga:


janko-m:
Thanks for the PR, but I don't see the point in adding Travis CI badges to READMEs. I will always run tests on all Ruby versions before I push to master, and if it is somehow happens that I don't, I will be notified via email.

PR #10, opened 2015-11-18
Support additional S3 fields

hakunin:
This allows us to pass in additional fields from the official doc, while maintaining server-side set size range.

Very useful if you want to set a redirect for example, you can add to your params:

```
fields: {
  success_action_redirect: 'http://example.org/<my_success_callback>'
}
```

http://docs.aws.amazon.com/AmazonS3/latest/dev/HTTPPOSTForms.html

### DIFF
--- a/lib/shrine/plugins/direct_upload.rb
+++ b/lib/shrine/plugins/direct_upload.rb
@@ -179,6 +179,7 @@ class Shrine
               r.get "presign" do
                 location = SecureRandom.hex(30).to_s + r.params["extension"].to_s
                 options = {}
+                options.merge!(r.params["fields"].symbolize_keys) if r.params["fields"]
                 options[:content_length_range] = 0..max_size if max_size
                 options[:content_type] = r.params["content_type"] if r.params["content_type"]
 


janko-m:
Thanks for the patch, but the problem with this approach is that it opens a security risk. For example, a user could pass the `acl` option with a potentially unwanted permissions. Or, they would currently even be able to pass the `:key` option, which would override the default key, and they would be able to upload a file anywhere to S3, which could cause problems.

I saw your question on Google Group, and I'm currently working a solution where you would be able to pass options as a configuration option when initializing the plugin. It would look something like this:

```rb
plugin :direct_upload, presign: ->(request) do # yields a Roda request object
  {success_action_redirect: "..."}
end
```

The idea is that you're yielded the `request` object, so that you have the flexibility to pass different options depending on the request. Would that work for you?

hakunin:
Yeah, thats perfect, thanks!

janko-m:
Pushed the commit to master: https://github.com/janko-m/shrine/commit/3b1af85a2cf972b594a7bcfbab8c982686b54293

I will be releasing version 1.0 soon (maybe today, but definitely in a few days), but until then you can pull from GitHub directly. Note that there are some smaller backwards incompatible changes (e.g. `delete_invalid` plugin was removed), I noted most of them in CHANGELOG.md, but I don't think it touches common usage so it should be safe to upgrade.

PR #12, opened 2015-11-20
Doc improvements

deivid-rodriguez:
Hi! Just wanted to say thanks for this gem. All previous solutions regarding uploading files seem overly complicated now. :+1: 

I read the README and tried out the stuff and notice some differences, minor errors & formatting issues. Feel free to cherry-pick any commits you like!

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,8 @@ records. In Shrine we do this by generating and including "attachment" modules.
 Firstly we need to assign the special `:cache` and `:store` storages:
 
 ```rb
+require "shrine/storage/file_system"
+
 Shrine.storages = {
   cache: Shrine::Storage::FileSystem.new("public", subdirectory: "uploads/cache"),
   store: Shrine::Storage::FileSystem.new("public", subdirectory: "uploads/store"),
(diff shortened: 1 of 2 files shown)


janko-m:
Thanks for the patches! I pushed https://github.com/janko-m/shrine/commit/a43b009602edb8ad05689ed706fdfd3ab5cc0dd5 which contains all the changes that I wanted.

* Good catch with `avatar_data` being a string, it's good to be precise in the README.
* I wanted to keep comments being in the same line so that it's more concise, especially now after GitHub's redesign when README is wider.
* In the meanwhile I removed the `:max_size` option from direct_upload. It's strange that I didn't add a commit description for that commit. The reason was that this can still be bypassed when uploading the file with the form (without direct upload), since the cached file won't be deleted. The real solution to this problem is to set a maximum request body size on the frontend server (nginx, apache). A Ruby solution is to use the `hooks` plugin and do something like this:

   ```rb
  class MyUploader < Shrine
    plugin :hooks
    def before_upload
    end
  end
   ```

deivid-rodriguez:
Thanks!

PR #17, opened 2015-11-24
Minor improvements

deivid-rodriguez:
A couple of minor improvements.

### DIFF
--- a/lib/shrine/plugins/pretty_location.rb
+++ b/lib/shrine/plugins/pretty_location.rb
@@ -22,7 +22,7 @@ class Shrine
           basename = "#{context[:version]}-#{basename}" if context[:version]
           original = dirname + slash + basename
 
-          [type, id, name, original].compact.join("/")
+          File.join(*[type, id, name, original].compact.map(&:to_s))
         end
       end
     end
--- a/shrine.gemspec
+++ b/shrine.gemspec
@@ -18,6 +18,7 @@ Gem::Specification.new do |gem|
 
   gem.add_dependency "down", ">= 1.0.3"
 
+  gem.add_development_dependency "rake"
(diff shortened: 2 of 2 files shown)


janko-m:
Thanks for the patches, but I'm not sure about the motivation behind them. The reason why I didn't set Rake as development dependency is that prefixing commands with `bundle exec` is for me around 3 times slower than running it normally and having `require "bundler/setup"`. But if Rake is set in the Gemfile and you don't prefix `rake` command with `bundle exec`, you would get version mismatches which can only be solved by updating Rake, which is a bit annoying to do every time you update the Rake gem locally outside of the Shrine project.

About the second change, it's arguably longer and more complex than the current code, can the current fail in some way? If not, I would prefer the simplicity.

deivid-rodriguez:
> But if Rake is set in the Gemfile and you don't prefix rake command with bundle exec, you would get version mismatches which can only be solved by updating Rake, which is a bit annoying to do every time you update the Rake gem locally outside of the Shrine project.

Wouldn't that happen with the other dependencies too if you don't use `bundler`? I added this because I'm used to prefix commands with `bundle exec` and I was getting errors when doing it with `rake`.

Regarding the second change, I'm not sure whether it could fail or not but I've seen this used in libraries where compatibility across different platforms is important (rubygems), and since I chose to do it this way in my own `pretty_location` method, I thought of proposing it here too.

janko-m:
Aha, I understand the `File.join` change now, because on Windows the separator is "\". The problem is that this change would on Windows generate "foo\bar\baz" for S3 as well, which is an invalid object key. I would accept a solution on FileSystem storage directly, which would replace all "/" with "\".

You're right about the Rake dependency, I cherry picked your commit.

deivid-rodriguez:
Thanks!

PR #18, opened 2015-11-24
Ruby 2.3 bug

deivid-rodriguez:
I run into this issue when running my app using `2.3.0-preview1`. Only the last block of the file would be copied. Not sure why it happens but it seems that `IO.copy_stream(io, filename)`  no longer works as expected. So I changed it to `IO.copy_stream(io, io)`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,15 @@ sudo: false
 rvm:
   - "2.1"
   - "2.2"
+  - "ruby-head"
   - "jruby-9.0.4.0"
+  - "jruby-head"
   - "rbx-2.5.8"
+
+matrix:
+  allow_failures:
+    - rvm: ruby-head
+    - rvm: jruby-head
 
 script: bundle exec rake test
 
--- a/test/file_system_test.rb
+++ b/test/file_system_test.rb
@@ -61,6 +61,12 @@ describe Shrine::Storage::FileSystem do
       assert_equal "image", input.read
     end
 
+    it "copies full file content" do
+      @storage.upload(input = fakeio("A" * 20_000), "foo.jpg")
+
+      assert_equal 20_000, @storage.open("foo.jpg").size
+   end
+
     it "sets file permissions" do
       @storage = file_system(root, permissions: 0755)
       @storage.upload(fakeio, "foo.jpg")


janko-m:
Thanks for the patch. However, this seems to be a [bug](https://bugs.ruby-lang.org/issues/11199) in Ruby trunk, which will hopefully get fixed once 2.3.0 is out. The bug is with the second argument being a Pathname, if it's a String then things work. While I could just call `#to_s`, I don't like the general idea of making workarounds for preview Ruby versions.

I don't think it makes much sense to run "head" Ruby versions in the Travis build. The reason is that they will usually have bugs until the stable version is released, and I don't think that there is any change that should be done in Shrine in that case, because the goal is to support only stable versions. I'm aware that you added them to "allowed failures", but it slows down the Travis build and I'm not sure what are the benefits.

deivid-rodriguez:
> Thanks for the patch. However, this seems to be a bug in Ruby trunk, which will hopefully get fixed once 2.3.0 is out. The bug is with the second argument being a Pathname, if it's a String then things work. While I could just call #to_s, I don't like the general idea of making workarounds for preview Ruby versions.

I totally agree, nice catch finding that bug upstream.

> I don't think it makes much sense to run "head" Ruby versions in the Travis build. The reason is that they will usually have bugs until the stable version is released, and I don't think that there is any change that should be done in Shrine in that case, because the goal is to support only stable versions. I'm aware that you added them to "allowed failures", but it slows down the Travis build and I'm not sure what are the benefits.

I think testing against "head" rubies helps spotting bugs in them (like this one) and this benefits the Ruby community overall because it helps building higher quality Ruby releases. Furthermore, it also works the other way around sometimes: bugs fixed in unreleased revisions of `ruby-core` reveal actual bugs in your code. Not this case, but happens too.

I totally understand that sometimes this can lead to very slow builds, but your test suite takes less than 1 second on my machine, I don't think this is one of those.

deivid-rodriguez:
Also, in the case of backwards incompatible releases, it helps you with the migration path.

janko-m:
Hmm, good point, it would be nice to help out the Ruby language this way. Is it possible to also add `rbx-head`, does this exist? I see now that `ruby-head` is erroring, is it something we can fix?

The thing that would be slowing down the build the most would be downloading rubies, not the actual test suite runtime.

Could you then rebase to master and include only the Travis changes?

deivid-rodriguez:
You're right about the slow down being caused by the download/installation of rubies. I think is acceptable though.

I rebased and added `rbx-head`. Let's see what happens!

deivid-rodriguez:
Ok, `rbx-head` is not there. I removed it back. And the `ruby-head` failure I've no idea what that is...

Also, what do you think about keeping the test I added? I think that specific functionality was not tested so it might make sense to add it. I could also add a remainder to the `ruby-core` issue and link it to the build failure and (maybe) it'll get fixed before 2.3 if the developers see that it causes real problems in libraries out there.

janko-m:
Yeah, it makes sense to keep the test itself :+1:. And we'll just keep the `ruby-head`, the Travis or Ruby core team will probably fix it at some point, and it's in "allowed failures" anyway.

deivid-rodriguez:
I re-added the test!

janko-m:
Awesome, thanks!

deivid-rodriguez:
Thanks to you! :)

deivid-rodriguez:
Wow, they've already fixed it! :smile: Next time you push something we'll confirm this is fixed.

janko-m:
It doesn't seem to have been fixed yet (https://travis-ci.org/janko-m/shrine/builds/93222848), but probably it will at some point.

PR #21, opened 2015-11-26
Pretty location generates shorter file IDs

gshaw:
Generate a 10 character alpha numeric string for cleaner file URLs.

Did not update the docs because the example in the plugin already shows it generating a 10 character suffix.


janko-m:
Thanks for the patch, looks good, I'll merge shortly.

PR #22, opened 2015-11-26
Provide S3 options for CloudFront

gshaw:
Specify Cache Control and ACL options for S3 uploads so that CloudFront can server the files as expected.

I'm not sure if this is the best way to accomplish this but I was only able to server uploaded assets over CloudFront by setting those two options.

More than happy to refactor this if given some guidance.  Especially as to how to test this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## HEAD
 
+* Add `cache_control` and `acl` parameters to S3 storage for CloudFront (gshaw)
+
 * Improve Windows compatibility in the FileSystem storage (janko-m)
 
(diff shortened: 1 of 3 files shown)


janko-m:
Thanks for the patch, it revealed a much deeper feature that was missing. In the commit I made this general, so that you can also pass some other options like `:server_side_encryption`. I also made it apply to presigning as well, since otherwise a lot of options need to be duplicated for presign (or direct uploads wouldn't work properly).

gshaw:
Thanks!  I wasn't sure if I was missing something or if it was missing so I hacked in something to solve my specific problem. Your fix is more general so thanks! :cookie:

PR #23, opened 2015-12-03
changed the presign call to use object(id) instead of @bucket

Chocksy:
- it allows the usage of custom endpoints for presigning files


janko-m:
Thanks for the patch!

PR #28, opened 2015-12-27
Add contribute insts to README plus 2 files to .gitignore

jasnow:
* Added contribute how-to instructions to README. (These were the ones I need to set up your repo). You are welcome to change/improve them.
 * Added 2 files to .gitignore file so others are not forced to use them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,7 @@ pkg/
 .sass-cache
 www/build
 coverage/
+
+# Added these to not force others to use them.
+.rvmrc
+.overcommit.yml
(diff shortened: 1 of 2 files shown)


janko-m:
Thanks for the PR, however, I don't agree with most of the changes.

For `.gitignore`, .rvmrc shouldn't be ignored in Shrine, because it isn't specific to this project (I'm using rbenv, for example), you should rather add it to your [global gitignore](https://help.github.com/articles/ignoring-files/). The same goes for .overcommit.yml, I don't even know what that is for :)

I would rather not have part of contributing instructions in README, and then have them repeat in CONTRIBUTING. I think developers who want contributing instructions should notice CONTRIBUTING.md, as it is the standard file for these instructions. But I didn't emphasize dependencies enough, so I'll change that.

jasnow:
Thanks for your response.

PR #31, opened 2016-01-10
Security fix: Plugins::ValidationHelpers#regex. ^$ are dangerous.

xzo:


janko-m:
Thanks for the patch!

PR #32, opened 2016-01-10
Fix ValidationHelpers docs: bad examples

xzo:


janko-m:
Thanks for the patch!

PR #35, opened 2016-01-25
Change Lotus into Hanami

katafrakt:
Lotus [has changed its name to Hanami](http://hanamirb.org/blog/2016/01/22/lotus-is-now-hanami.html) and so did my gem and example repository. This PR fixes URLs and names.

### DIFF
--- a/index.html
+++ b/index.html
@@ -563,8 +563,8 @@
           <ul>
             
             <li>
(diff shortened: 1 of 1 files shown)


janko-m:
Thanks for the update!

Issue #36, opened 2016-01-28
Moving files on S3 from cache to store


mkcode:
I have my application setup and working very nicely with direct to s3 uploads using shrine.

Looking into the `delete_uploaded` plugin, I notice that it does not support `UploadedFile` and therefore will not operate on S3 cache to S3 store.

I notice the `move_to` method on `aws-sdk`: http://docs.aws.amazon.com/sdkforruby/api/Aws/S3/Object.html#move_to-instance_method

Have you evaluated this at all? I would think that we can get the functionality of `delete_upload` on s3 to s3 copying if we use this `move_to` method.

But also, I notice `move_to` just calls `copy_to` and then `delete`, leading me to think that all the copy operations are synchronous an therefore ok to use?

Have you evaluated this at all? If it is possible, do you think this should be part of the `delete_uploaded` plugin or a separate plugin?

janko-m:
Yeah, I have gave this topic a long thought, it took me a long time to reach a decision, and am sometimes still reevaluating this choice but always arrive to the same conclusion.

The reason why I don't allow deleting `UploadedFile` in the `delete_uploaded` plugin is because cached files in general shouldn't be deleted on upload, but rather the cache storage should be cleared periodically in a recurring task. To explain the reason for this, consider the following scenario:

```rb
class ImageUploader < Shrine
  def process(io, context)
    if context[:phase] == :store
      tempfile = io.download
      original = io # cached file (instance of Shrine::UploadedFile)
      medium = resize_to_limit(tempfile, 500, 500)
      small = resize_to_limit(tempfile, 300, 300)

      {original: original, medium: medium, small: small}
    end
  end
end
```

I find this scenario to be very common. If the `delete_uploaded` plugin would delete `Shrine::UploadedFile` instances, storing the above hash of versions would first upload the `:original` cached file (and delete it), and then upload `:medium` and `:small`. This means that if you're doing backgrounding, the cached file that was potentially in use (for graceful degradation until the background job finishes) would get deleted *before* storing finished and the record is updated, which would momentarily result in a broken link on the cached file.

This may not be a big deal in low-traffic applications, but the bottom line is that it would make your app sometimes look broken, and the bug is very subtle (you wouldn't get an error report for it), so I wanted to ensure that it can never happen with Shrine (unless you're using the `moving` plugin, but it's not intended to be used with versions anyway).

[...]

PR #37, opened 2016-01-31
Promote hooks

mkcode:
Hi @janko-m,

After thinking about your response to #36, I came up with a (hopefully) acceptable solution. This PR adds hooks to the `promote` method.

After this PR, you may write an uploader like follows:

```ruby
class AttachmentUploader < Shrine
  plugin :backgrounding
  plugin :hooks
  plugin :versions, names: [:original, :thumb]

  def process(io, context)
    if context[:phase] == :store
      thumb = resize_to_limit!(io.download, 350, 350)
      {original: io, thumb: thumb}
    end
  end

  def after_promote(cached_file, record:)
    cached_file.delete
  end
end

```

The `after_promote` hook will run after the `store` versions have been uploaded and the new value is persisted to the DB. Deleting the cache file here leaves 0 time when the store and DB are in an inconsistent state. I hope you find this an acceptable solution to all the issues discussed in #36. Let me know if I can improve this PR in anyway. Thanks for your time!

### DIFF
--- a/lib/shrine/plugins/hooks.rb
+++ b/lib/shrine/plugins/hooks.rb
@@ -71,6 +71,69 @@ class Shrine
     # In that case you should always somehow mark this key as private (for
     # example with an underscore) so that it doesn't clash with any
     # existing keys.
+    #
+    # Promote hooks are triggered when `promote` or `finalize` method is run on
+    # your record's attacher instance or the `Shine::Attacher` class in the case
+    # that the `:backgrounding` plugin is used. Similar to all previous hooks,
+    # they are executed in the following order:
+    #
+    # * `around_promote`
+    #     * `before_promote`
+    #     * PROMOTE
+    #     * `after_promote`
+    #
+    # Promote hooks are called with 2 arguments, `cached_file` and `record`.
+    # `cached_file` is the `cache` file upload. `record` is the full instance of
+    # the associated record. You should always call `super` when overriding a
+    # hook, as other plugins may be using hooks internally, and without `super`
+    # those wouldn't get executed.
+    #
+    # The promote hooks are the only shrine hooks that execute around an update
+    # to the record and its underlying database. In the case of `ActiveRecord`,
+    # `before_promote` and `after_promote` operate when ActiveRecord's
+    # `before_save` and `after_commit` hooks are run. This also means that the
+    # `record` argument supplied to `before_promote` and `after_promote` will
+    # have its uploader in the cache state and store state respectively:
+    #
+    #     In this example; `ImageUploader` is attached to `image` on the record
+    #     and the uses the `versions` plugin.
+    #
+    #     class ImageUploader < Shrine
+    #       def before_promote(cached_file, record:)
+    #         record.image_data #=> {
+    #           "id": "43kewit94.jpg",
+    #           "storage": "cache",
+    #           "metadata": { "size": 384393, ... }
+    #         }
+    #         super
+    #       end
+    #     end
+    #
+    #     class ImageUploader < Shrine
+    #       def after_promote(cached_file, record:)
+    #         cached_file #=> {
+    #           "id": "43kewit94.jpg",
+    #           "storage": "cache",
+    #           "metadata": { "size": 384393, ... }
+    #         }
+    #         record.image_data #=> {
+    #           "original": {
+    #             "id": "43kewit94-original.jpg",
+    #             "storage": "store",
+    #             "metadata": { "size": 114917, ...}
+    #             }
+    #           },
+    #           "thumb": {
+    #             "id": "43kewit94-thumb.jpg",
+    #             "storage": "store",
+    #             "metadata": { "size": 11493, ...}
+    #           }
+    #         super
+    #       end
+    #     end
+    #
+    # Similar to the other hooks, you can realize some form of communication
+    # between the hooks by sending more arguments to the rest argument.
     module Hooks
       module InstanceMethods
         def upload(io, context = {})
--- a/lib/shrine/plugins/hooks.rb
+++ b/lib/shrine/plugins/hooks.rb
@@ -151,6 +151,27 @@ class Shrine
 
         def after_delete(*)
         end
+
+
+        def around_promote(*args)
+          before_promote(*args)
+          result = yield
+          after_promote(*args)
+        end
+
+        def before_promote(*)
+        end
+
+        def after_promote(*)
+        end
+      end
+
+      module AttacherMethods
+        def promote(cached_file)
+          result = nil
+          store.around_promote(cached_file, record: record) { result = super }
+          result
+        end
       end
     end
 
--- a/lib/shrine/plugins/hooks.rb
+++ b/lib/shrine/plugins/hooks.rb
@@ -107,9 +107,7 @@ class Shrine
     #         }
     #         super
     #       end
-    #     end
     #
-    #     class ImageUploader < Shrine
     #       def after_promote(cached_file, record:)
     #         cached_file #=> {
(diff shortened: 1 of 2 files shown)


janko-m:
First of all, thank you for making the PR with tests included.

However, adding promote hooks to the uploader class isn't as intuitive as it might appear. The reason is that the `Shrine` class itself does't have the concept of "promoting". It only knows how to upload (and optionally process) and delete files, the `Shrine::Attacher` class is the one that is in charge of these "higher-level" concepts. That's why in my opinion these hooks don't belong in `Shrine`, and moving them to the attacher class would probably be a bit confusing (since then two kinds of hooks would exist).

I was thinking about adding some kind of promote hooks in the past, but I couldn't quite fit them in anywhere.

If it's really not convenient to set up some automatic cache clearing, you could achieve the same behaviour with the `module_include` plugin (didn't test it):

```rb
class MyUploader < Shrine
  plugin :module_include
  attacher_module do
    def promote(cached_file)
      result = super
      delete!(cached_file, phase: nil)
      result
    end
  end
end
```

Come to think about it, it would probably be nice to make a plugin for this. The main reason this is not in the base is because it could cause errors in combination with the `moving` plugin (because it's not possible to determine whether the cached file was moved, without saving some state). Would that work for you?

mkcode:
> I was thinking about adding some kind of promote hooks in the past, but I couldn't quite fit them in anywhere.

Right. I put the hooks in the uploader so that there can be `Uploader` specific behavior. Putting them in the `Attacher` class would not allow for this. In my opinion, for the common user  (one that does not deeply inspect the source code), the `Uploader` is where all behavior customizations are made, and so it makes sense to put it here, but I agree that from a code SRP perspective, this is mixing responsibilities.

> If it's really not convenient to set up some automatic cache clearing

My issue with this (and I suspect others as well) is that this is offloading application logic outside of the application. This is made worse when (like my current organization), there is a separate team / people in charge of managing AWS resources. I would rather not co-ordinate this across different servers, and especially across different people.

> you could achieve the same behaviour with the module_include plugin

Ah ha! I had missed this. Yes, this essentially the same thing as the promote hooks, but better because it is clear that the code is executing in a different context.

> Come to think about it, it would probably be nice to make a plugin for this. The main reason this is not in the base is because it could cause errors in combination with the moving plugin (because it's not possible to determine whether the cached file was moved, without saving some state). Would that work for you?

Yes this would work. Maybe it would be possible to extend the `moving` plugin with this functionality to prevent these errors. Where we override the promote function if both the `cache` and `store` storage are S3?

Thanks again for the thoughtful response. :sparkles:

janko-m:
> I put the hooks in the uploader so that there can be `Uploader` specific behavior. Putting them in the `Attacher` class would not allow for this.

Actually, in order for the plugin system to respect inheritance, each `Shrine` subclass gets its own "copy" of the `Attacher` class, so any overriding of that class would be uploader-specific. The module_include plugin also relies on this.

> My issue with this (and I suspect others as well) is that this is offloading application logic outside of the application.

I agree with this point, there should probably be a way to make cached files automatically deleted (safely), this was previously part of base functionality until I noticed issues with the moving plugin. I did have a workaround then, but it was an ugly hack.

However, the reason why I didn't put much priority to this is because even with this in place, you still have to use the remove_invalid plugin to be safe if you're doing normal caching. But if you're doing direct uploads, then it's not possible to prevent the cache directory from filling up, because the user can upload a new file before the form is submitted, leaving the old file in the "nowhere land".

So even if I added this, with direct uploads you still need the automatic cache clearing. Because of that I'm not sure if it's worthwhile to add this behaviour. What do you think?

mkcode:
> Actually, in order for the plugin system to respect inheritance, each Shrine subclass gets its own "copy" of the Attacher class, so any overriding of that class would be uploader-specific

Good to know :+1: . Thanks.

> But if you're doing direct uploads, then it's not possible to prevent the cache directory from filling up, because the user can reupload files before the form is submitted, leaving the unused files in the "nowhere land".

It it possible to make multiple POST requests to the same s3 `presigned_post`. If you reuse the `presigned_post` response in the frontend and do not use the special `${filename}` as part of the key, later uploads will overwrite the previous uploads and will not cause unnecessary objects in the cache. This has been working for me in my application. This would prevent a lot, but there are still some cases, such as if the user does not submit the form at all where the cache would fill up. Though, you could catch all the cases by queuing an image delete job in the far future to delete the cache image when the `presigned_post` is generated.

> So even if I added this, with direct uploads you still need the automatic cache clearing. Because of that I'm not sure if it's worthwhile to add this behavior. What do you think?

The `module_include` plugin, will work fine for me. Thanks for that info. So it's whatever you think is best for the project. Perhaps the best thing would be to simply update the example application. :-)

janko-m:
> This would prevent a lot, but there are still some cases, such as if the user does not submit the form at all where the cache would fill up. Though, you could catch all the cases by queuing an image delete job in the far future to delete the cache image when the presigned_post is generated.

Cool, I didn't have your trick in mind. However, as you mentioned, the user can just abandon the form. I just think it's never a good idea to have a hole which can be abused. Especially if you're accepting multiple file uploads (like in the example application).

The solution that is arguably the simplest and most effective is to setup automatic cache clearing (especially for S3), so when I think about I'm not anymore for the idea of adding a plugin for deleting cached file. People can use the `module_include` snippet if they really want. Thank you for raising this issue and the discussion.

mkcode:
Thanks again for being so helpful.

In my never ending quest to not leak this app behavior, I wrote and am now using the following plugin in my application. I'm sure that you can see what I am going by looking at it. This should cover all cases of clearing the cache, as `Shrine::Attacher.delete_presigned_post_later(s3_post, 1.week.from_now)` is run on every presigning request. Let me know if you think this method is worth exploring for Shrine, and I can put together a PR. It could be added to the `direct_upload` plugin as well.

```ruby
class Shrine
  module Plugins
    module DeletePresignedPostLater
      OPTS_KEY = :backgrounding_delete_presigned_post_later

      module AttacherClassMethods
        def self.load_dependencies(uploader, *)
          uploader.plugin :backgrounding
        end

        def delete_presigned_post_later(presigned_post = nil, wait_until = nil, &block)
          if block
            shrine_class.opts[OPTS_KEY] = block
          else
            key_id = presigned_post.fields['key']
            data = {
              'delete_presigned_post' => true,
              'key_id' => key_id
            }

            background_block = shrine_class.opts[OPTS_KEY]
            wait_until ||= 1.week.from_now

            instance_exec(data, wait_until, &background_block)
          end
        end

        def delete(data = nil, &block)
          if data && data['delete_presigned_post'] == true
            key_id = data['key_id']

            uploader = shrine_class.new(:store)
            uploader.storage.delete(key_id)
          else
            super(data, &block)
          end
        end
      end
    end

    register_plugin(:delete_presigned_post_later, DeletePresignedPostLater)
  end
end
```

janko-m:
Interesting idea. First I thought that it would be better to create a more generalized version of this plugin, one which would delete any cached file, not just direct uploaded ones. But then I remembered that direct S3 uploads don't touch the app.

I think this plugin is too specific to be shipped with Shrine, it's only for direct presigned uploads, and it wouldn't handle all cases; for example, if the key uses a dynamic `${filename}` like you mentioned, there is no way to know the key until the file is uploaded.

I'm not against a generic plugin that would make all cached files automatically deletable in a user-specified background job, but at the moment I don't know how it would look like, and if it's even possible to implement it in a simple and robust way. Since `cache` will almost always be either FileSystem or S3, both for which it's simple to setup automatic cache clearing, I don't think this plugin should ship with Shrine. If you decide to create an external plugin, I will gladly link to it on the website.

PR #41, opened 2016-02-06
Avoid opening an UploadedFile just in order to close it

reidab:
I'm using Shrine to manage direct S3 uploads and was seeing an unexpected GET to S3 during the `after_commit` phase. This came to light when writing some tests using [webmock](https://github.com/bblimke/webmock) to stub all external network requests.

In the case of an S3 `UploadedFile` that's been initialized from JSON with no processing step required, the copy happily happens on the server and `UploadedFile#io` never needs to be populated. But after the copy completes, the `ensure` block of `UploadedFile#copy` calls `UploadedFile#close`, which references the unpopulated `UploadedFile#io` and triggers `UploadedFile#open`.

### DIFF
--- a/lib/shrine.rb
+++ b/lib/shrine.rb
@@ -717,8 +717,10 @@ class Shrine
         # Part of Shrine::UploadedFile's complying to the IO interface.  It
         # delegates to the internally downloaded file.
         def close
-          io.close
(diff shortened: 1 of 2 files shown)


janko-m:
Great catch! Yes, this is definitely the desired behaviour, thank you for the fix. :heart:

PR #47, opened 2016-02-25
Pretty location plugin with namespaced models

kaapa:
When pretty location generates a location for a namespaced model it should replace '::' with '/'. If this is not done the S3 storage's URI method (given host option is set) attempts to use `URI.join` which does not produce an expected result when the path contains '::'.

```
2.2.3 :001 > require 'uri'
 => true
2.2.3 :002 > URI.join('https://example.com', 'namespaced::openstruct/something.jpg').to_s
 => "namespaced::openstruct/something.jpg"
2.2.3 :003 > URI.join('https://example.com', 'namespaced/openstruct/something.jpg').to_s
 => "https://example.com/namespaced/openstruct/something.jpg"
```

Besides this one could argue that the directory structure should follow the namespacing of the models.

This PR is bit problematic in the sense that for filesystem storage it introduces a breaking change. Then again for the S3 storage it does not break things as the previous implementation simply did not work.

Finally, I'd like to say thank you for this feature packed and yet understandable codebase. I'm new to this library so I may have understood something wrong. The inspiration for the PR came from a real world use case where I had to override the `generate_location` method for an uploader with S3 storage and host option set, attached to a namespaced model so I believe I'm onto something here. Perhaps there's another way around this and as previously stated this PR has its downsides as it can break a functioning default pretty location setup when the storage is filesystem.

### DIFF
--- a/lib/shrine/plugins/pretty_location.rb
+++ b/lib/shrine/plugins/pretty_location.rb
@@ -14,7 +14,7 @@ class Shrine
     module PrettyLocation
       module InstanceMethods
         def generate_location(io, context)
-          type = context[:record].class.name.downcase if context[:record] && context[:record].class.name
+          type = context[:record].class.name.downcase.gsub('::', '/') if context[:record] && context[:record].class.name
           id   = context[:record].id if context[:record].respond_to?(:id)
           name = context[:name]
 
--- a/test/plugin/pretty_location_test.rb
+++ b/test/plugin/pretty_location_test.rb
@@ -2,14 +2,18 @@ require "test_helper"
 require "ostruct"
 
 describe "the pretty_location plugin" do
+  module NameSpaced
+    class OpenStruct < ::OpenStruct; end
+  end
+
   before do
     @uploader = uploader { plugin :pretty_location }
   end
 
   it "uses context to build the directory" do
-    uploaded_file = @uploader.upload(fakeio, record: OpenStruct.new(id: 123), name: :avatar)
+    uploaded_file = @uploader.upload(fakeio, record: NameSpaced::OpenStruct.new(id: 123), name: :avatar)
 
-    assert_match %r{^openstruct/123/avatar/[\w-]+$}, uploaded_file.id
+    assert_match %r{^namespaced/openstruct/123/avatar/[\w-]+$}, uploaded_file.id
   end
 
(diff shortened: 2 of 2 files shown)


kaapa:
A non-breaking alternative solution appears to be prepending the path with a slash.

```
2.2.3 :001 > require 'uri'
 => true
2.2.3 :002 > URI.join('https://exaple.com', 'namespaced::openstruct/something.jpg').to_s
 => "namespaced::openstruct/something.jpg"
2.2.3 :003 > URI.join('https://exaple.com', '/namespaced::openstruct/something.jpg').to_s
 => "https://exaple.com/namespaced::openstruct/something.jpg"
```

Personally I'd prefer the solution given in this PR, but I can understand that this alternative fix would be backwards compatible and hence a smoother release. Happy to hear your thoughts on the topic!

janko-m:
Good catch, thanks for the patch!

I think whenever people namespace database models, they usually provide a namespace like `MyApp::Models::User`, meaning that models will be put into the same namespace without any semantical meaning, so I don't think the namespace should be included in the location. The class names should be differently named in any case, because their names maps to a "flat" table. So I would like more if only the class name is used.

Btw, this change isn't backwards incompatible, because when you change how the location is generated the old files will still continue to work, since the location is written and read from the column, so `generate_location` isn't called anymore once the file is uploaded.

kaapa:
Thank you for clarifying the bit about backwards compatibility!

I guess this is a case where one simply can not please all the use cases by one solution. Eg. in my Rails application the models are namespaced like `Article::Translation` and `BusinessPlan::Translation` etc. in which case cutting the path to just the inner class name would be counter intuitive.

One other option would be to flatten the namespace by replacing `::` with `_` instead of path separator, but this of course comes down to which approach Shrine considers as the usual model namespacing convention. I do understand that if the models are namespaced by application name then prefixing all the paths with that app's name + word model would not be ideal - except in a IMO silly edge case where one would store multiple applications' files in a single S3 bucket.

Naturally, I'm OK with overriding my uploaders' `generate_location` methods if you'd rather see the "inner class" solution implemented. Also I'm happy to take this PR to any direction you see as the best fit.

kaapa:
Great solution - thank you! :)

janko-m:
You're welcome, and thank you for the discussion :)

PR #49, opened 2016-02-26
Rewind IO cursor after failed image dimension extraction by FastImage

kaapa:
I made this PR in bit of haste so please review accordingly.

FastImage gem does not rewind IO cursor if the dimension extraction fails eg. due to providing a non-image file. Non rewinded IO causes S3 upload to fail with Aws::S3::Errors::BadDigest before file can be validated.

### DIFF
--- a/lib/shrine/plugins/store_dimensions.rb
+++ b/lib/shrine/plugins/store_dimensions.rb
@@ -62,7 +62,10 @@ class Shrine
         private
 
         def _extract_dimensions_with_fastimage(io)
-          FastImage.size(io)
+          FastImage.size(io, raise_on_failure: true)
+        rescue FastImage::FastImageException
+          io.rewind
+          nil
         end
       end
 
--- a/test/fixtures/invalid_image.jpg
+++ b/test/fixtures/invalid_image.jpg
@@ -1,0 +1,1 @@
+Invalid image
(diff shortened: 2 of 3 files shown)


kaapa:
I just noticed the bit about image bombs in the readme. I don't really understand how FastImage protects from those so I guess swallowing all `FastImageException`s could break that protection. Perhaps we should only rescue `UnknownImageType` exceptions?

janko-m:
Thank you very much for the PR. Actually, rescuing all `FastImageException` is exactly what we want, because we don't want to react on any errors, since the type/extension validations are checking if the file is an image. As for image bombs, I think that Fastimage doesn't abort but still manages to extract dimensions in this case.

I think this is an important fix, but I think it would be much better if this was fixed in Fastimage itself, so I sent a PR there: https://github.com/sdsykes/fastimage/pull/66.

kaapa:
Yeah, I was also thinking that a more fitting place for the fix would be FastImage itself, but did not have time to dive into the source. Lets hope they'll accept the change! :)

PS. I most likely still need to resort to monkey patching since I've had to lock FastImage to v1.8.1 as my project has a dependency conflict (with Addressable gem) starting from FastImage v1.9.

janko-m:
Hmm, that's unfortunate. Sam Saffron opened a pull request for removing the Addressable dependency, but it's still waiting. Apparently the only reason why Addressable is used over the URI standard library is because there was some bug concerning underscores in URLs, which was fixed in Ruby 2.2. But that can surely be fixed somehow inside Fastimage, I mean, everyone uses URI standard library for parsing all the time.

But I don't want only people who update Fastimage to get this fixed in Shrine (when/if the PR eventually gets merged), so it's better to merge your fix.

Issue #44, opened 2016-02-17
Mongoid Support?


nitinstp23:
Hey,

Does `shrine` supports `mongoid` as an ORM.

If not, would you consider a PR?

janko-m:
Shrine doesn't support Mongoid at the moment. It is still on my TODO list, but since I'm not using MongoDB I never got to it. It would be great to have Mongoid integration, though I would prefer if it was a separate gem, just because I think MongoDB generally isn't as essential as SQL databases, but mostly because it's better that it's maintained by a person who uses it :smiley:. If you decide to create it, let me know so that I can link to it on the website (or send me a PR).

nitinstp23:
Cool :+1:

PR #50, opened 2016-02-29
Added support for Mongoid - #44

nitinstp23:
* Added `mongoid` ORM.
* Added `presign` method to `FileSystem` storage.
* Don't disable the default POST route if `presign` is true.
* Pass the POST endpoint's URL parameter to storage's `presign` method.

### DIFF
--- a/lib/shrine/plugins/backgrounding.rb
+++ b/lib/shrine/plugins/backgrounding.rb
@@ -102,7 +102,7 @@ class Shrine
           if background_promote = shrine_class.opts[:backgrounding_promote]
             data = {
               "uploaded_file" => get.to_json,
-              "record"        => [record.class.to_s, record.id],
+              "record"        => [record.class.to_s, record.id.to_s],
               "attachment"    => name.to_s,
             }
 
@@ -120,7 +120,7 @@ class Shrine
           if background_delete = shrine_class.opts[:backgrounding_delete]
             data = {
               "uploaded_file" => uploaded_file.to_json,
-              "record"        => [record.class.to_s, record.id],
+              "record"        => [record.class.to_s, record.id.to_s],
               "attachment"    => name.to_s,
               "phase"         => phase.to_s,
             }
--- a/lib/shrine/plugins/direct_upload.rb
+++ b/lib/shrine/plugins/direct_upload.rb
@@ -95,7 +95,7 @@ class Shrine
     #
     # While Shrine only accepts cached attachments on form submits (for security
     # reasons), you can use this endpoint to upload files to any storage, just
-    # add it do allowed storages:
+    # add it to allowed storages:
     #
     #     plugin :direct_upload, allowed_storages: [:cache, :store]
     #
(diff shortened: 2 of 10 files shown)


janko-m:
Wow, that's a lot of changes, thank you very much! Do you think we could split them up into separate PRs? But let's first discuss them.

## 1. Typos

Always welcome, it would be great to have a separate commit/PR for them.

## 2. Converting record IDs to strings in backgrounding

What does this change fix? It's just that I can imagine someone using an ORM/database without proper typecasting (e.g. something improvised), so maybe it could happen that now when the ID is a string, finding could fail. I mean, probably everybody has proper typecasting, I would just like to know if current behaviour here was causing any problems.

## 3. Specifying location with `key` param

This one I'm not so sure it's a good idea. Mainly, I don't think it's secure, because it allows the user/attacker to upload the file anywhere in the cache. I mean, I don't see direct negative implications from it at the moment, but I would rather not leave this possibility. Is there a reason you want to upload the file to a specific location? Could that location be generated inside `Shrine#generate_location`? If not, maybe we could make your behaviour opt-in somehow.

## 4. Easy presign testing

I'm having this one in my TODO list, but didn't get to it yet, because I didn't managed to figure out the best way to do this yet in a secure way. If users are using presigns (passed the `:presign` option), I don't think the "direct upload" route should always turned on. This is because it performs a potentially expensive operation (uploading to external storage), which is fine if storage is filesytem but not if it's S3. And likely the developer doesn't want to have this route enabled if they're using presigns.

But I do like the idea of "fake" presigns. What I think is a good solution is, if the storage supports presigns (responds to `#presign`), then we disable the "direct upload" route, otherwise we leave it enabled. I think this is better than `FileSystem` storage implementing `#presign`, and having to delete the `:url` option in S3 storage (maybe aws-sdk decides to add `:url` option to presigns).

What do you think about it?

## 5. Mongoid support

It's awesome that you've implemented Mongoid support. However, as I said before, I would really like this to be a separate gem (which I would link on the website). Just because I don't consider MongoDB as essential as SQL databases here, and as a separate gem it would be maintained by a person who actually uses MongoDB.

I see that you've locked Mongoid to versions before 5.x. Is there some specific reason for it? Just because lower version of Mongoid use Moped, which I assume will get deprecated over the offical [Mongo Ruby Driver](https://github.com/mongodb/mongo-ruby-driver), which is the one that Mongoid 5.x uses.

nitinstp23:
## 1. Typos
Cool will send a separate PR for typos.

## 2. Converting record IDs to strings in backgrounding
This was happening due to a bug in the way I was monkey patching `BSON::ObjectId` to convert it to json. Finally got a solution here - https://github.com/rails-api/active_model_serializers/issues/354#issuecomment-55717441
So this change in the `backgrounding` plugin is not required anymore. :+1: 

## 3. Specifying location with `key` param
The reason for doing this is that the `presign` request generates the filename using this - 
```ruby
location = SecureRandom.hex(30) + request.params["extension"].to_s
```
But the `:post` endpoint doesn't know about this generated `filename` and tries to generate a new `filename` here - https://github.com/janko-m/shrine/blob/master/lib/shrine.rb#L314
Which fails to find the uploaded file in the `cache` directory.
I did this fix in a bit of exigency. Do let me know if there is a better way to fix this.

## 4. Easy presign testing
I tried to implement this in a bit of exigency, but do let me know your thoughts on this. BTW I didn't understood the idea of `fake` presigns :smile:

## 5. Mongoid support
Cool. I'll extract the `mongoid` plugin to a separate gem and will let you know to add the link on the website.

janko-m:
Ok, I completely understand know why you added the "key" param. It indeed makes sense that the user is able to choose the location when direct uploading, I was probably a bit paranoid thinking that it wasn't secure. I would accept a (separate) PR for this.

By "fake" presigns I meant that for a storage we're conforming to the interface of presigns, even though that particular storage (FileSystem) in reality doesn't have presigns. And we're doing it mainly for easier testing.

Ok, great, I think we can close this now, and you can just split these changes into multiple PRs.

janko-m:
To follow-up, I've [implemented](https://github.com/janko-m/shrine/commit/a653e1a94118f9e47d95cd3d20b1f8ab910201f4) the ability to use presigns with FileSystem storage (or any other) for testing, the implementation is similar to yours. I've also fixed the typos.

nitinstp23:
Thanks @janko-m  :+1:

PR #61, opened 2016-05-05
Fix file command in determine_mime_type plugin

zaeleus:
The `file` utility's option parser is greedy and interprets every
argument that begins with a hyphen as an option. This means pathnames
that start with a hyphen cause the command to fail, either by invalid
options or never passing it required input. For example,
`file -image.jpg` looks like combined options and is parsed as `-i`, `-m`,
`-a` (invalid), etc.

To prevent this, `--` is appended to the command to signify the end of
options and allows subsequent arguments to be excluded from option
parsing.

---

Here's [a standalone script](https://gist.github.com/zaeleus/5784e0b966555c5d008ec534ad7b1501) that shows the incorrect behavior.

````
$ wget -O -image.jpg http://i.imgur.com/aDNltvP.jpg
$ wget https://gist.githubusercontent.com/zaeleus/5784e0b966555c5d008ec534ad7b1501/raw/c311f5b098e91eecbe6dc814df57c5bfc839121f/shrine_61.rb
$ ruby shrine_61.rb
Usage: file [-bchikLNnprsvz0] [-e test] [-f namefile] [-F separator] [-m magicfiles] [-M magicfiles] file...
       file -C -m magicfiles
Try `file --help' for more information.
["isn't of allowed type: [\"image/jpeg\"]"]
````

### DIFF
--- a/lib/shrine/plugins/determine_mime_type.rb
+++ b/lib/shrine/plugins/determine_mime_type.rb
@@ -86,7 +86,7 @@ class Shrine
         # if it's a file, because even though the utility accepts standard
         # input, it would mean that we have to read the whole file in memory.
         def _extract_mime_type_with_file(io)
-          cmd = ["file", "--mime-type", "--brief"]
+          cmd = ["file", "--mime-type", "--brief", "--"]
 
           if io.respond_to?(:path)
             mime_type, _ = Open3.capture2(*cmd, io.path)
--- a/test/plugin/determine_mime_type_test.rb
+++ b/test/plugin/determine_mime_type_test.rb
@@ -16,6 +16,24 @@ describe "the determine_mime_type plugin" do
       mime_type = @uploader.send(:extract_mime_type, fakeio(image.read))
       assert_equal "image/jpeg", mime_type
     end
+
+    describe "when the path looks like an option" do
+      let(:path) { "-image.jpg" }
+
+      before do
+        FileUtils.cp(image, path)
+      end
+
+      after do
+        FileUtils.rm(path)
+      end
+
+      it "does not fail and determines content type from file contents" do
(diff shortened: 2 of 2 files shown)


zaeleus:
This is perhaps another issue and seeing https://travis-ci.org/janko-m/shrine/builds/128177438 fail, the current file command differs in behavior following symlinks on Linux (does not follow) and OS X (follows). In the context of shrine, it may be more intuitive to follow symlinks using the `--dereference` option.

janko-m:
Thank you for the patch! I'm curious to why your file is a symlink in this case, while the other `text/fixtures/image.jpg` isn't. Probably has something to do with changing the directory.

I would probably makes sense to follow symlinks in general. I'm not seeing a use case for it at the moment, but not seeing any downside either. I will add it to my TODO list, to evaluate if this is secure, and will see about adding it.

Do you think that instead of changing the directory, you could just copy `spec/fixtures/image.jpg` into project root (with the filename of `-image.jpg`), and then simply delete it after the test has finished?

zaeleus:
> I'm curious to why your file is a symlink in this case, while the other text/fixtures/image.jpg isn't.

My first patch had a symlink named [`-image.jpg` that pointed to `image.jpg`](https://github.com/zaeleus/shrine/blob/bd93a569fd45f1fcecab66fb4c2a9a2156459053/test/fixtures/-image.jpg), but after seeing it fail on Travis, I pushed a new one that used a copy instead, which confirmed my suspicion that `file` on Linux doesn't follow symlinks by default like on OS X.

> Do you think that instead of changing the directory, you could just copy spec/fixtures/image.jpg into project root (with the filename of -image.jpg), and then simply delete it after the test has finished?

I updated the test to do this.

PR #73, opened 2016-08-03
Improve README

yogodoshi:
First: awesome readme, very detailed, as is the documentation.

Im suggesting some small changes to improve the readme because of some problems/ doubts I had while following it. I think could facilitate for more people that could have the same doubts/ errors :)


janko-m:
Thank you!

PR #79, opened 2016-08-28
Add missing link of FineUploader in docs

willnet:
current http://shrinerb.com/rdoc/files/doc/direct_s3_md.html

image


janko-m:
Thanks for the fix!

PR #84, opened 2016-09-27
pretty_location optional id_partition extracted from paperclip

ttp:
Added `id_partition` option to `pretty_location` to be able to shard files between subfolders.

```rb
Shrine.plugin :pretty_location, id_partition: true
```
```rb
photo = Photo.create(image: File.open("nature.jpg"))
photo.image.id #=> "photo/000/000/034/image/34krtreds2df.jpg"
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -580,6 +580,14 @@ photo = Photo.create(image: File.open("nature.jpg"))
 photo.image.id #=> "photo/34/image/34krtreds2df.jpg"
 ```
 
+```rb
+Shrine.plugin :pretty_location, id_partition: true
+```
+```rb
+photo = Photo.create(image: File.open("nature.jpg"))
+photo.image.id #=> "photo/000/000/034/image/34krtreds2df.jpg"
+```
+
 If you want to generate locations on your own, you can override
 `Shrine#generate_location`:
 
--- a/lib/shrine/plugins/pretty_location.rb
+++ b/lib/shrine/plugins/pretty_location.rb
@@ -22,16 +22,26 @@ class Shrine
     #
     #     plugin :pretty_location, namespace: "/"
     #     # "blog/user/.../493g82jf23.jpg"
+    #
+    # To spread files across multiple subfolders based on id partitioning pass
+    # :id_partition option:
+    #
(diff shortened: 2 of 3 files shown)


janko-m:
@ttp Thanks for the PR! It would definitely be good to have a way to partition the locations, both for filesystem and [S3](http://docs.aws.amazon.com/AmazonS3/latest/dev/request-rate-perf-considerations.html). I will take a look at this more closely by the end of the week.

mma7:
http://shrinerb.com/rdoc/classes/Shrine/Plugins/PrettyLocation.html
Can you please update the docs!
Thanks.

ttp:
> Can you please update the docs!

@mma7 done

janko-m:
After reading [this Amazon article](http://docs.aws.amazon.com/AmazonS3/latest/dev/request-rate-perf-considerations.html), it looks that there are many ways to apply partitioning of folders, each good for a specific use case.

With that in mind, I don't think it makes sense that Shrine ships with one specific way of partitioning. In general I would encourage users to write their own partitioning scheme, which will work best for their use case, since it depends on what you want to partition on (id, extension, or something else), the rate of GET requests, how the particular storage handles partitioning etc.

While I wouldn't like to include this behaviour in Shrine core, you can still release it as an external plugin if you find it useful (and I think it makes more sense that way than for it to be part of the pretty_location plugin). If you decide to make it, I would gladly add it to the website.

ttp:
> After reading [this Amazon article](http://docs.aws.amazon.com/AmazonS3/latest/dev/request-rate-perf-considerations.html)

@janko-m wondering how to create such hash key prefix as we have cache/ & store/ prefixes for all files. Probably should move them into separate buckets or just remove prefix setting for store in s3?

janko-m:
I was hoping that S3 partitions from left to right, that it doesn't take only the top-level "folder", but that it continues partitioning on what differs. But maybe that's not how it works.

In any case, not using prefixes should be easy. It might even make sense to leave the prefix for `cache/`, and for main storage have it on the top-level. However, removing prefixes when there are already files uploaded is difficult and requires more than 1 stage of deployment, like any data migration.

PR #86, opened 2016-09-29
minor code comment fix

ericduvon:


janko-m:
@ericduvon Thanks for the patch!

PR #88, opened 2016-09-29
fix broken link for Using Attacher guide

ericduvon:
I think this is the correct link you wanted as the rdoc link goes to 404 page


janko-m:
@ericduvon Thanks for the patch, but this is actually the intended link, it's just not present yet because I haven't deployed the guide to the website (which I do on each version release). The reason why I want to link to the website instead of GitHub is so that it always represents the latest released version of Shrine (and not changes which haven't been released yet).

PR #89, opened 2016-09-30
Redesign the website

silvenon:
Please check if the changes made to `_data/external.yml` are ok with you, I wasn't sure if the previous location of *Plugins* was purely cosmetic.

Changes:

- It's using the native OS font (on macOS it's San Francisco, on Android it's Droid Sans etc.), so visitors won't have to download web fonts and the web page will feel more native.
- Improved responsiveness, now it's about as responsive as it can be.
- Uploading animation in the header along with a paperclip icon, communicating *attachment*.
- New color scheme, better text contrast, prettier button.
- Fixed `<dl>` markup, now there's a single `<dl>` per title.
- Add a simple favicon.


janko-m:
@silvenon Thank you, brother, it looks awesome!! There are just two small changes that I would like to make:

* I would prefer to keep "Storages" and "Plugins" sections one under the other, like in the original website (because they compose all external gems)
* I would also like to keep the "README" link inverted (with a red background), like it was on the original website. I think it nicely emphasizes the main link.

silvenon:
Fixed. I suggest moving `/www` to `/docs` so you can get rid of gh-pages ([more info](https://help.github.com/articles/configuring-a-publishing-source-for-github-pages/)).

janko-m:
@silvenon Thanks!!

Thanks for the info about `gh-pages`, I will definitely look into it in the future. At the moment it doesn't really fit as I already have a `/doc` folder. And I'm not sure if this would mean that the website is deployed every time I push to the repo (which isn't the desired behaviour, I need to be able to control it manually).

PR #91, opened 2016-10-05
Preserve filename on download

jonasheinrich:
Also set the filename in `Content-Disposition` header on forced download links.

### DIFF
--- a/lib/shrine.rb
+++ b/lib/shrine.rb
@@ -763,6 +763,7 @@ class Shrine
 
         # Calls `#url` on the storage, forwarding any options.
         def url(**options)
+          options.merge!(shrine_metadata: metadata)
           storage.url(id, **options)
         end
 
--- a/lib/shrine/storage/s3.rb
+++ b/lib/shrine/storage/s3.rb
@@ -237,8 +237,20 @@ class Shrine
       #
       # [`Aws::S3::Object#presigned_url`]: http://docs.aws.amazon.com/sdkforruby/api/Aws/S3/Object.html#presigned_url-instance_method
       # [`Aws::S3::Object#public_url`]: http://docs.aws.amazon.com/sdkforruby/api/Aws/S3/Object.html#public_url-instance_method
-      def url(id, download: nil, public: nil, host: self.host, **options)
-        options[:response_content_disposition] = "attachment" if download
+      def url(id,
+              download: nil,
+              public: nil,
+              host: self.host,
+              shrine_metadata: nil,
+              **options)
+        options[:response_content_disposition] = if download
+          if shrine_metadata &&
+              filename = shrine_metadata['filename']
+            "attachment; filename=#{filename}"
+          else
+            "attachment"
+          end
+        end
 
         if public
           url = object(id).public_url(**options)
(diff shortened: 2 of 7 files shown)


janko-m:
Thanks for the PR!

There is actually a similar commit on master, https://github.com/janko-m/shrine/commit/61ce15e69f510fe53ea1b8b87d2a191a57d00313, which sets the `Content-Disposition` header on upload itself using the original filename. However, I see now that this actually won't work for forced download links, as we're overriding `:response_content_disposition` with just `attachment`, so the filename is lost.

The proposed change unfortunately isn't ideal, as passing `:shrine_metadata` to every storage would break some storages like [shrine-cloudinary](https://github.com/janko-m/shrine-cloudinary/blob/2f1cc79d390ccca9a4740c4b1cb60c8ae0f33ef1/lib/shrine/storage/cloudinary.rb#L71-L73). We should probably search for alternative solution.

One that currently comes to mind is to set the correct `Content-Disposition` dynamically on upload using the `upload_options` plugin:

```rb
plugin :upload_options, store: ->(io, context) do
  filename = context[:metadata]["filename"]
  {content_disposition: "attachment; filename=#{filename.inspect}"}
end
```

That means that all objects with that `Content-Disposition` would have forced download links by default, I don't know if that's a downside for some use cases.

jonasheinrich:
You're welcome :) And sorry I didn't write a pr description, just realised that. 

Yes, I've seen that this change breaks a lot when fixing the tests. 

In our application we'd like to have both, `inline` and `attachment` and both with the filename. So the proposed solution with `upload_options` won't work for us. 

One workaround would be to call `Attacher#url` with the desired option. But that would somehow validate the single responsibility principle so I would like to avoid it. 

I guess adapting the interface for the other storages' `url` methods isn't a valid way to go? This would probably break storages that you don't control the code of. 

I can't think of any other solution right now. Somehow we need to pass the metadata into the method but we can't change the interface nor put it into `options`. Tricky.

janko-m:
I see, that totally makes sense, thanks for pointing that out. Another solution that could work would be to modify the `default_url_options` plugin so that it can take a black, to allow creating dynamic URL options:

```rb
Shrine.plugin :default_url_options, store: ->(io, **options) do
  if options.key?(:download)
    {response_content_disposition: "attachment; filename=#{io.original_filename.inspect}"}
  else
    {response_content_disposition: "inline; filename=#{io.original_filename.inspect}"}
  end
end
```

Would that be feasible?

jonasheinrich:
Thanks a lot for taking so much time. 

That seems to be a good solution. I don't need the `inline` part because I already use the commit you mentioned above. But the `attachment` part seems great. Gonna give it a try tomorrow.

janko-m:
@jonasheinrich Ok, I pushed https://github.com/janko-m/shrine/commit/bb86043e32435e83e811faa403c60a18e1a48b22 which adds this feature to the `default_url_options` plugin.

jonasheinrich:
@janko-m Thanks a lot for your effort. 

I just had to make a small change, see https://github.com/janko-m/shrine/pull/92. This only fixes the S3 storage though.

PR #92, opened 2016-10-06
Don't overwrite option if already set

jonasheinrich:
Referring to https://github.com/janko-m/shrine/pull/91


janko-m:
Great, thanks!

PR #94, opened 2016-10-07
Add support postgres json type for activerecord

donbobka:
Add support of postgres json and jsonb fields for activerecord plugin.

### DIFF
--- a/lib/shrine/plugins/activerecord.rb
+++ b/lib/shrine/plugins/activerecord.rb
@@ -112,6 +112,17 @@ class Shrine
       end
 
       module AttacherMethods
+        # Support for Postgres JSON columns.
+        def _set(uploaded_file)
+          data = case record.class.columns_hash["#{name}_data"].type
+                 when :json, :jsonb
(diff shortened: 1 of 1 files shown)


janko-m:
@donbobka Thanks for the patch!

From my own testing, it seems that the issue is that ActiveRecord requires assigning values using a Hash, and that assigning a JSON String like Shrine doesn't work because ActiveRecord treats it as a single value of a JSON type string, right?

I'm assuming that this patch works for you, but if that's the case I think we would be relying on too much ActiveRecord magic. Because `uploaded_file` object that is being written to the column will be either a `Shrine::UploadedFile` object, or a hash of `Shrine::UploadedFile` objects. While ActiveRecord might serialize them properly now (maybe through `#as_json`), I think we should be explicit and assign a Hash directly.

I will push a fix for it shortly.

donbobka:
@janko-m Thanks for response and fix. I thought about deserialization from JSON, but this is a little bit strange. Check my PR #95 .

PR #95, opened 2016-10-08
Extract serialization logic

donbobka:
This allows to change serialization logic by overiding Attacher#serialize_data method.

### DIFF
--- a/lib/shrine.rb
+++ b/lib/shrine.rb
@@ -596,7 +596,7 @@ class Shrine
         # It returns nil if the value is blank.
         def read
           value = record.send(data_attribute)
-          value unless value.nil? || value.empty?
+          (value.nil? || value.empty?) ? nil : convert_data_read(value)
         end
 
         # Uploads the file using the #cache uploader, passing the #context.
@@ -657,15 +657,28 @@ class Shrine
           shrine_class.opts[:validate]
         end
 
-        # Dumps the UploadedFile to JSON format and writes the result to the
-        # attachment attribute.
-        def _set(uploaded_file)
-          write(uploaded_file ? uploaded_file.to_json : nil)
+        # Converts the UploadedFile with method `convert_data_write`
+        # and writes the result to the attachment attribute.
+        def _set(value)
+          write(value ? convert_data_write(value) : nil)
         end
 
         # Writes to the `<attachment>_data` attribute on the model instance.
         def write(value)
           record.send(:"#{data_attribute}=", value)
+        end
+
+        # Converts value from data attribute to accepted format
+        # for method `uploaded_file`
+        def convert_data_read(value)
+          value
+        end
+
+        # Converts hash representation of UploadedFile or UploadedFile
+        # for writing to data attribute
+        # `value` is a Hash or UploadedFile
+        def convert_data_write(value)
+          value.to_json
         end
 
         # Temporary method used for transitioning from :phase to :action.
--- a/lib/shrine/plugins/activerecord.rb
+++ b/lib/shrine/plugins/activerecord.rb
@@ -120,13 +120,12 @@ class Shrine
           record.save(validate: false)
         end
 
-        # Support for Postgres JSON and JSONB columns.
-        def write(value)
-          if activerecord_json_column?
-            value = JSON.parse(value) if value
-          end
+        def convert_data_write(value)
+          activerecord_json_column? ? value.as_json : value.to_json
+        end
 
-          super(value)
+        def convert_data_read(value)
+          activerecord_json_column? ? value.as_json : JSON.parse(value)
         end
 
         def activerecord_json_column?
--- a/lib/shrine/plugins/sequel.rb
+++ b/lib/shrine/plugins/sequel.rb
@@ -111,11 +111,12 @@ class Shrine
       end
 
       module AttacherMethods
-        # Support for Postgres JSON and JSONB columns.
-        def read
-          value = super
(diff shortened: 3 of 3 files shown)


janko-m:
Thanks, I like this improvement! I will test this out and merge tomorrow.

janko-m:
@donbobka Ok, I looked at this more closely now. I definitely like the idea of extracting this into separate method, and removing the unnecessary dumping-parsing from JSON.

However, I think that the name "serialization" isn't entirely accurate here. What we're doing is just converting the value to a format that can be assigned to the data attribute, and then the data attribute (which comes from the ORM) is the one that's performing actual serialization (before saving to the database).

I think a good name for this is simply "conversion", so to name the method `#convert_data`. But I would actually like to generalize this. The `sequel` plugin also supports Postgres JSON columns, and while it doesn't need to convert the value on assignment (the column still accepts JSON string values), on reading it needs to convert the Sequel-specific Hash-like object to an actual hash.

So I think it would make sense to create a plugin which acts as a layer around writing to and reading from the data attribute. The ActiveRecord and Sequel plugins would then depend on this `attribute_conversion` plugin (a name that I came up with now), where the ActiveRecord plugin would hook up only to the "writing" part, and Sequel plugin to the "reading" part.

What do you think?

donbobka:
@janko-m I agree with you that it's not about serialization, it's about preparing data for storing in database.

I have an idea that all conversion logic should be extracted totally to Data Layer plugins (sequel, activerecord, ...). Each data layer plugin responsible for conversion of input hash.

janko-m:
> I have an idea that all conversion logic should be extracted totally to Data Layer plugins (sequel, activerecord, ...). Each data layer plugin responsible for conversion of input hash.

So, my sugesstion was to have a generic `attribute_conversion` plugin, which the "data layer" plugins would be able to use to perform specific conversions that they need.

But there is another use case, even if you're using the "data layer" plugin, I think people might still not always want to save the data in the same format. For example, I was thinking that if you're using Mongoid, you might want to save the data as a Hash instead of a JSON string, because then I think you can query it. There could be other use cases, I think.

Therefore, I think this `attribute_conversion` plugin would still make sense, because people might want to use it directly. In that case the plugin would probably override `Attacher#_set` and `Attacher#read` (like activerecord and sequel plugins are doing now), so the base functionality wouldn't have to know about this feature. How does that sound?

donbobka:
What sort of code will be in `attribute_conversion` plugin?  
As I understand your idea: put default shrine converting logic(with json string) to `attribute_conversion` plugin.

`attribute_conversion` plugin:
```ruby

def convert_data_read(value)
  JSON.parse(value)
end

# `value` is a Hash
def convert_data_write(value)
  value.to_json
end
```

`activerecord` plugin:
```ruby
def convert_data_write(value)
  activerecord_json_column? ? value: super(value)
end

def convert_data_read(value)
  activerecord_json_column? ? value : super(value)
end
```

`sequel` plugin:
```ruby
def convert_data_read(value)
  value.respond_to?(:to_hash) ? value.to_hash : super(value)
end
```

donbobka:
From my point of view, it's impossible to use shrine without `attribute_conversion` plugin. If so, why we extract it to plugin?

donbobka:
Changed my PR.

janko-m:
@donbobka Thank you for this discussion, it's really valuable to get a second opinion!

> From my point of view, it's impossible to use shrine without attribute_conversion plugin. If so, why we extract it to a plugin?

I was thinking that the core behaviour would still be the same (serialization and deserialization from JSON), but that the plugin would just hook in front of it like you did in the first version of the PR.

Note that for PORO models we cannot make the change to assign a Hash instead of a JSON string, because that would break backwards compatibility. While I see the arguments for using a hash, I like JSON serialization because it is the same as Shrine already accepts uploaded files in `Attacher#assign`, and also `#to_json` can recursively serialize hash of versions (so we don't need to make special case when the value is an `UploadedFile` or a hash of `UploadedFile` objects).

Also, `#as_json` is an ActiveSupport-specific method, so it cannot be in the base functionality (though it can be in ActiveRecord plugin, because ActiveRecord already loads ActiveSupport). You noticed that `Shrine::UploadedFile` implements `#as_json`, but that's actually only to prevent loading ActiveSupport from breaking Shrine, because without that ActiveSupport breaks recursive `#to_json` for hash of versions.

> And I guess that conversion logic can be a very different for each data layer gem. That's why each data layer plugin should choose a best way to convert hash by itself.

I agree with this wholeheartedly, and I really like your overview of different "data layers". You're right, In my Mongoid example the corresponding `mongoid` plugin could automatically choose the write a value in different format, depending on if the field type is "string" or an "object".

I'm definitely convinced now that an `attribute_conversion` plugin wouldn't be useful, at least for any use cases that I can think of, but the extraction of `#convert_data_write` and `#convert_data_read` is definitely useful, and makes it more straightforward to implement "data layer" plugins.

I will merge your PR shortly, but just make a modification to still keep assigning JSON strings to POROs for the argument that I explained above. Thank you!

donbobka:
I returned back PORO conversion logic and removed `as_json` from core.

janko-m:
@donbobka Thanks! Since it's a bit late here, I will merge this tomorrow!

janko-m:
@donbobka I've made a few modifications in https://github.com/janko-m/shrine/commit/e2d52a622e582a6d81e4ce76fb386828eb3b654a.

donbobka:
@janko-m Thanks 👍

PR #102, opened 2016-11-03
Add dot to file extension when missing

jonasheinrich:
Currently, you have to request the presigned POST with something like `...?extension=.jpg`. It would be nice to be able to drop the dot, i.e. `...?extension=jpg`.


janko-m:
Thanks for the PR! I agree with the enhancement, merging it shortly 👍

PR #104, opened 2016-11-04
Validate extensions case insensitive

jonasheinrich:
If we have `jpg` in the whitelist it would make sense to accept `JPG` as well.


janko-m:
@jonasheinrich Thanks for the patch! Looks good, merging shortly.

PR #105, opened 2016-11-04
Overwrite existing metadata on S3

jonasheinrich:
This is especially useful for `Content-Type`. S3 sets this to `application/octet-stream` by default. So when you use the `direct_upload` plugin (and don't let the presign endpoint take the mime type) files uploaded directly to the cache in S3 will always have this content type. 
The AWS S3 cli does not overwrite metadata by default, only when told so via `metadata_directive`.  This is what this PR aims to do. 

But if you set `metadata_directive` to `'REPLACE'` no metadata is copied at all. So you need to fetch the old metadata and merge it into the new metadata. 

Unfortunately, it seems like `metadata_directive` influences the behaviour when updating e.g. `content_type`, but `Object#metadata` only includes metadata of the form `x-amz-meta-*`. So we might loose some metadata like `content_encoding`. I'm not completely sure how to solve this yet.

### DIFF
--- a/lib/shrine/storage/s3.rb
+++ b/lib/shrine/storage/s3.rb
@@ -302,6 +302,13 @@ class Shrine
 
       # Copies an existing S3 object to a new location.
       def copy(io, id, **options)
+        options[:metadata_directive] = 'REPLACE'
+        existing_metadata = object(io.id).metadata
+        options[:metadata] ||= {}
+        options[:metadata].merge!(existing_metadata) { |_, new_value, _|
+          new_value
+        }
+
         options = {multipart_copy: true, content_length: io.size}.update(options) if multipart?(io)
         object(id).copy_from(io.storage.object(io.id), **options)
       end


janko-m:
I was trying to write a script to test which attributes/headers get copied over by default, and which don't get copied over with `metadata_directive: "REPLACE"`, but gave up in the middle because there are so many options 😛 

However, I think having this option by default could lead to surprising behaviour. For example, someone might want to pass `:content_diposition` when generating a presign, and then expect that it will automatically be copied over on S3 copy.

Before S3 copy we *could* just fetch the existing object attributes, and then merge them in with the new ones (like `:content_type`) and pass `metadata_directive: "REPLACE"`. However:

* Fetching this information requires an additional HEAD request, which I wouldn't want to add, especially for people who are not using backgrounding. In this PR the `Object#metadata` will be blank, because we don't have this information without making a HEAD request.
* When the S3 object attributes are fetched, they don't map 1-to-1 with the options for the S3 copy request, so it would be additional work to transform them.

The `content_type` case can be easily solved by deriving it from the extension, and passing it in to the presign:

```rb
plugin :direct_uploads, presign_options: -> (request) do
  extension = request.params["extension"]
  content_type = Rack::Mime.mime_type(extension)
  {content_type: content_type}
end
```

Note that `Rack::Mime.mime_type` expects extension to contain a dot (`.jpg`), so if you're passing it without a dot you'll need to include it for the method.

I would like to make this the default behaviour (to derive `content_type` from the extension), but the S3 storage isn't the only one which has presigns, [Cloudinary](https://github.com/janko-m/shrine-cloudinary/blob/c899875b935a45bc322a5e18be9c2132ebeecb4d/lib/shrine/storage/cloudinary.rb#L83) and [Uploadcare](https://github.com/janko-m/shrine-uploadcare/blob/c167516ba3002f7c880bfdea2e349731e7a7dddd/lib/shrine/storage/uploadcare.rb#L74), and they accept different presign options.

janko-m:
I was also thinking about adding it to the `S3#storage` itself, that it automatically generates a `:content_type` based on the extension from the passed `id`. However, I cannot use `Rack::Mime.mime_type` there as it is Rack-specific, and Shrine doesn't have to necessarily be used with Rack.

We could theoretically do that only if Rack is defined:

```rb
def presign(id, **options)
  if defined?(Rack)
    extension = File.extname(id)
    content_type = Rack::Mime.mime_type(extension)
    options = options.merge(content_type: content_type)
  end
  # ...
end
```

But I don't think I like the idea of having web-framework specific logic inside storage classes, it doesn't feel right.

janko-m:
I will close this pull request, as I don't think that's it's good idea to have `metadata_directive: "REPLACE"` on by default.

ndbroadbent:
+1 for having `metadata_directive: "REPLACE"` on by default for S3. I just ran into this problem with direct uploads, and it took me a few hours to debug this issue, until I found this thread. I'll add the following code for now:

```ruby
plugin :upload_options, store: -> (io, context) do
    { metadata_directive: "REPLACE" } if io.is_a?(Shrine::UploadedFile)
  end
```

But it would have been great if Shrine could just override the content type automatically, instead of preserving `binary/octet-stream` from the direct upload.

janko-m:
@ndbroadbent I added to my TODO list to use `metadata_directive: "REPLACE"` in Shrine 3 by default, I agree that it's a more sensible default.

I instinctively wanted to change it already in the next 2.x version, but some users like @jonasheinrich might be relying on the current behaviour, so I would still prefer to keep backwards compatibility in 2.x. For now I [updated](https://github.com/janko-m/shrine/commit/b3c6e5a762dffaff5a3ab6dcb6c83cf75ef67743) the documentation explaining the default S3 behaviour and using `:metadata_directive`.

PR #114, opened 2016-12-05
A little test cleanup

hms:
* Added what looks like a missing test
* Fixed all assert_equal nil depreciations
   Changed all assert_equals nil, something to assert_nil something
* Added rubymine and rvm files to .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@ pkg/
 .sass-cache
 www/build
 coverage/
+.rvmrc
+.idea
--- a/test/attacher_test.rb
+++ b/test/attacher_test.rb
@@ -38,13 +38,13 @@ describe Shrine::Attacher do
     it "rejects stored files for security reasons" do
       stored_file = @attacher.store!(fakeio)
       @attacher.assign(stored_file.to_json)
-      assert_equal nil, @attacher.get
+      assert_nil nil, @attacher.get
     end
 
     it "accepts nils" do
       @attacher.assign(fakeio)
       @attacher.assign(nil)
-      assert_equal nil, @attacher.get
+      assert_nil @attacher.get
     end
 
     it "ignores empty strings" do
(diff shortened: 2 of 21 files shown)


janko-m:
@hms Thanks for the PR!

1. About the `.gitignore`, I think that only files that have to do with that project should be gitignored locally, `.rvmrc` or `.idea` files should be added to global gitignore. I personally have a `~/.gitignore` file, and the following `~/.gitconfig` entry:

```
[core]
  # a global .gitignore
  excludesfile = ~/.gitignore
```

2. Oh, I almost forgot that `assert_equal nil, x` got deprecated in Minitest, I cherry-picked your commit with updates to `assert_nil`.

3. There already is a [test](https://github.com/janko-m/shrine/blob/f855aca252913259f168ebc75e0723c350ea4e3a/test/shrine_test.rb#L188-L191) for testing that returning nil in `#generate_location` raises a `Shrine::Error`. It's not contained under `describe "#generate_location"` because this check happens outside of that method.

hms:
@janko-m Thanks for the quick feedback.

#1.  I agree with the idea of project specific ignores.  But since Shrine is open source, and we want as many people to work on it as possible (/reasonable?), then the definition of 'files that have to do with the project' takes on a slightly different meaning.

On my laptop, I rely on rvm and in order to set up my fork and maintain my workflow, it was necessary to include an .rvmrc.  If I add .rvmrc to my global ignores, then I break other projects.  Rubymine has the same issues (.idea).

I would suggest for an open source project, common tools should be fair game for .gitignore, even if they are not your preferred tools.

#3.  Thanks for the catch.  I don't think I noticed the other test because the default test suite only runs the shrine file (which ties into my other issue :-) )

Hal

janko-m:
> I would suggest for an open source project, common tools should be fair game for .gitignore, even if they are not your preferred tools.

But isn't it easier that developers globally gitignore files that are related to their personal workflow, which they have to do only once, rather than having to gitignore these files on each project they work on? 😉

hms:
I misspoke above. You are correct that the .idea really should go into my global ignore (which I have, but haven't used in so long, I forgot about it).

But .rvmrc is another case altogether as I don't want to ignore it for some projects, just Shrine at the moment.

PR #125, opened 2017-01-09
Small text fix on validation_helpers.rb

brunowego:
👍


janko-m:
Thanks for the fix!

brunowego:
Thank you for this incredible gem!

PR #131, opened 2017-01-20
Update README.md

ckk-scratch:
Had to dig into additional docs to realize it was :sequel OR :activerecord


janko-m:
@ckk-scratch Thanks for pointing out an unclear part of the documentation. I prefer to use `or` instead of `||`, so I'll push that patch instead.

PR #135, opened 2017-01-29
Add validation helper to validate max filename length

rob-murray:
Hi,
Thanks for the great work on this gem.

This is one of the validations I have added the couple of times I have used the gem - maybe could be useful to have in the project?

I can change this to support minimum length or allow a range if you would like?

### DIFF
--- a/test/plugin/validation_helpers_test.rb
+++ b/test/plugin/validation_helpers_test.rb
@@ -32,6 +32,22 @@ describe Shrine::Plugins::ValidationHelpers do
       assert_equal true, @attacher.validate_min_size(1)
       assert_empty @attacher.errors
     end
+
+    it "uses the default error messages" do
+      @attacher.assign(fakeio)
+      @attacher.validate_min_size 2*1024*1024
+      assert_equal ["is too small (min is 2.0 MB)"], @attacher.errors
+    end
+
+    it "accepts custom error messages" do
+      @attacher.assign(fakeio)
+      @attacher.validate_min_size 2*1024*1024, message: ->(min) { "< #{min/1024/1024}" }
+      assert_equal ["< 2"], @attacher.errors
+
+      @attacher.assign(fakeio)
+      @attacher.validate_min_size 2*1024*1024, message: "is too small"
+      assert_equal ["is too small"], @attacher.errors
+    end
   end
 
   describe "#validate_max_width" do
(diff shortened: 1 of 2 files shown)


janko-m:
Thank you very much for the pull request! ❤️

Since Shrine only stores the filename to the database, and doesn't actually use it for uploading files (unlike CarrierWave or Paperclip), might I ask what is the use case for validating the filename?

Btw, don't pay attention to Travis, the new JSON gem broke the build, and I haven't managed to fix it yet.

rob-murray:
Yes, since Shrine don't have to fit the filename in a 255 character column, I'm not sure it's applicable to everyone. My case is a system that historically people have stored files with names having few hundreds of characters, this makes files hard to identify and display in a list so we try to stop people dumping files with crazy long names and think about labeling the file better.

janko-m:
@rob-murray Oh, that makes total sense. Ok, I'll give it a proper review today or tomorrow, and merge it if everything is good.

test/plugin/validation_helpers_test.rb
```diff
@@ -32,6 +32,22 @@
       assert_equal true, @attacher.validate_min_size(1)
       assert_empty @attacher.errors
     end
+
+    it "uses the default error messages" do
```
rob-murray: I rearranged these tests to fit within the describe block for the method.

janko-m:
@rob-murray Thanks again for the pull request, but after giving this some more thought, I don't think it's a good idea to do filename length validations from the UX perspective, therefore I wouldn't like to have it in Shrine.

All other validations (filesize, MIME, extension, dimensions) are related to the file itself, and the reason to do those validations is because we want that users can upload only certain kind of files.

However, the length of the filename doesn't say anything about the file, only how the file is named, and that can always be changed. If I personally would have the same kind of requirements, I wouldn't require users to reupload the file if the filename is too long, I would rather truncate the filename on the frontend.

If you still want to have filename validations, the easiest way to add them to your uploader using the `module_include` plugin to add validation methods to `Shrine::Attacher`:

```rb
class MyUploader < Shrine
  plugin :module_include

  attacher_module do
    def validate_filename_max_length(max, message: nil)
      get.original_filename.to_s.length <= max or add_error(:filename_max_length, message, max) && false
    end
  end
end
```

Thanks anyway for the pull request, I'm sorry it didn't get merged.

rob-murray:
No problem

PR #139, opened 2017-02-06
Improved backgrounding plugin documentation

rubyconvict:
If the backgrounding plugin is configured inside the uploader class, the following error occurs inside the job:
```
Shrine::Attacher.promote(data)
```
```
NoMethodError: undefined method `promote' for Shrine::Attacher:Class
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -781,6 +781,20 @@ class DeleteJob
 end
 ```
 
+If the plugin is configured in specific uploader class, exact attacher class should be used:
+```rb
+  def perform(data)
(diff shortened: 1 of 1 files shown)


janko-m:
@rubyconvict Thank you very much for the patch, I really appreciate when people contribute with documentation! ❤️

I added this to the `backgrounding` plugin documentation rather than the README, since the README is just supposed to show a general example. I also added an example how you can register promote/delete hooks from inside the uploader class.

PR #148, opened 2017-03-10
Updates documentation wording

danielpowell4:
This documentation is about migrating from Paperclip to Shrine.

In one sentence, however, documentation misspoke and referenced "CarrierWave" instead of "Paperclip"

This `itty-bitty` commit cleans this up


janko-m:
Thanks for the patch!

PR #156, opened 2017-04-13
Fix typo in Plugins::MetadataAttributes docs

Morozzzko:
This PR doesn't really propose any significant changes. At all. 

So I was reading docs on the [MetadataAttribbutes plugin](http://shrinerb.com/rdoc/classes/Shrine/Plugins/MetadataAttributes.html). 
The docs said ```plugin :metadata_values```. 
Decided to try it out... Yup. 

```
LoadError:
       cannot load such file -- shrine/plugins/metadata_values
```

Replacing `metadata_values` with `metadata_attributes` did the job.

Checked docs for other plugins (listed at shrinerb.org). Everything else seems okay.


janko-m:
Oops, thanks for the correction!

PR #167, opened 2017-06-07
Accept Attacher options via Attachment.new

ypresto:
I wrote below snippet to pass `store:` and `cache:` args to Attacher but it did not work.

```rb
include ImageUploader::Attachment.new(:file, store: :private_store, cache: :private_cache)
```

So I implemented it..!


janko-m:
Thanks, looks good! I will merge shortly, and add some minor tweaks.

PR #169, opened 2017-06-19
Validations inheritance

printercu:
Following https://groups.google.com/forum/#!topic/ruby-shrine/sN6Zdg1YJxI


janko-m:
Thanks, looks great! I'll take one more look tomorrow, and then merge it in.

PR #171, opened 2017-06-28
Fix mime_type on empty file causes TypeError or unexpected value

ypresto:
When empty file is passed to determine_mime_type plugin,

- `:file` analyzer will return `"cannot open: No such file or directory"` for mime type string.
- `:filemagic` will raise TypeError, which is caused by passing nil to `filemagic.buffer()`.

This PR fixes it..!

(filemagic one causes SEGV when gcc is used >_< https://github.com/blackwinter/ruby-filemagic/pull/24 )

### DIFF
--- a/lib/shrine/plugins/determine_mime_type.rb
+++ b/lib/shrine/plugins/determine_mime_type.rb
@@ -142,7 +142,12 @@ class Shrine
           require "open3"
 
           cmd = ["file", "--mime-type", "--brief", "-"]
-          options = {stdin_data: io.read(MAGIC_NUMBER), binmode: true}
+          buffer = io.read(MAGIC_NUMBER)
+          # Emulate behavior of file command:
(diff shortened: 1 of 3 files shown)


janko-m:
Thanks, looks good! I think we can make it simpler by calling `#to_s` on the result, so for `file` we don't need a new conditional. I'll make that change upon merging.

janko-m:
I changed the behaviour to always return `nil` on empty IO, regardless of the analyzer. Because `file` returns `application/octet-stream` and `filemagic` returns `application/x-empty` for empty IO, I wanted to make things consistent. I think it's right for MIME type to be blank for empty IOs, because there is literally no content to determine the MIME type from.

ypresto:
ceb66c753686095c07d7cb654a6d6e737b8464d8

Thanks!! :tada:

PR #175, opened 2017-07-19
Add missing location to module_eval

printercu:


janko-m:
Great catch, thank you!

PR #179, opened 2017-08-04
Fix styling issues following github-rubocop rules

manzan46:
Just started the process to move from Carrierwave to Shrine [@GoodGym](www.goodgym.org) (Great work btw :+1: ), and notice some old 1.9 ruby Syntax in the docs and code.

So I thought it could be nice to enforce new version ruby style.

One of the best way to to it is to use [rubocop](https://github.com/bbatsov/rubocop) and [github version of it](https://github.com/github/rubocop-github) which follow [Ruby community style guide](https://github.com/bbatsov/ruby-style-guide), and keeping this gem up-to-date will permit to stay in sync with the style when moving to newer version

One exception which I was unsure was `AllowAsExpressionSeparator` which was complaining about following example with inline semi column: 
`@uploader.class.add_metadata { |io, context| io.read; nil }` 
or 
`@storage.instance_eval { def upload(io, id, *); super; io.close; end }`

[Here](https://gist.github.com/manzan46/3372c14c6b3fc841989ea24b210c710f) is a list of all original complain of rubocop

Keep it up the good work

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -8,7 +8,7 @@ Rake::TestTask.new do |t|
   t.warning = false
 end
 
-task :default => :test
+task default: :test
 
 RDoc::Task.new do |t|
   t.rdoc_dir = "www/build/rdoc"
@@ -29,7 +29,7 @@ RDoc::Task.new do |t|
   ]
 end
 
-task :rdoc => "website:rdoc_github_links"
+task rdoc: "website:rdoc_github_links"
 
 namespace :website do
   task :build do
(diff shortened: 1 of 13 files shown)


janko-m:
@manzan46 Thanks for the patch!

I wouldn't like to add Rubocop to Shrine, as for me it has always been an annoying obstacle when contributing to other gems. However, I would like to pull in the style corrections from this PR. Would you mind updating your PR to include only style changes?

manzan46:
Hey @janko-m Sorry for the delay on this one, removed the gem :D

janko-m:
Thanks, cherry-picked to `master`.

PR #185, opened 2017-09-20
Custom attacher options

arg:
This was partially implemented in https://github.com/janko-m/shrine/pull/167, but `Attacher` class crashes if we pass any options other then `cache` and `store`.

```ruby
class Image < ActiveRecord::Base
  include Shrine[:asset, max_upload_size: 1.megabyte] # will crash with "unknown keyword: max_upload_size" message
  ...
end
```

### Why do we need this?

In our app there are few different models having the same `Shrine::Attachment` module included. The only attachment module would totally fit our needs but we'd like to have ability to customize it (e.g specify different validation options like maximum file size or so) in each model. It's not possible with the current Shrine implementation, we have to inherit several attachment classes from our base instead.

With the proposed improvement we will be able to have only one attachment class, customizing it for each model:

```ruby
class Image < ActiveRecord::Base
  include Attachment[:asset, max_file_size: 5.megabytes]
  ...
end

class User < ActiveRecord::Base
  include Attachment[:avatar, max_file_size: 1.megabyte]
  ...
end

class Attachment < Shrine
  ...
  Attacher.validate do
    validate_max_size(options[:max_file_size])
  end
  ...
end
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## HEAD
 
+* Allow specifying custom attacher options (@arg)
+
 * Modify `UploadedFile#extension` to always return the extension in lowercase format (@janko-m)
 
 * Downcase the original file extension when generating an upload location (@janko-m)
--- a/doc/attacher.md
+++ b/doc/attacher.md
@@ -27,10 +27,11 @@ attacher.get                                          # equivalent to `photo.ima
 The attacher object exposes the objects it uses:
 
 ```rb
-attacher.record #=> #<Photo>
-attacher.name   #=> :image
-attacher.cache  #=> #<ImageUploader @storage_key=:cache>
-attacher.store  #=> #<ImageUploader @storage_key=:store>
+attacher.record  #=> #<Photo>
+attacher.name    #=> :image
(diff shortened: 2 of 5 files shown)


janko-m:
Thanks for the pull request!

However, this already _is_ possible with the current implementation. The `Attacher.validate` block in which validations are defined is not a DSL, it's just plain Ruby block which is evaluated in the context of a `Shrine::Attacher` instance, which has access to the model instance and attachment name (this is documented in the README).

```rb
Attacher.validate do
  record # #<Image> or #<User>
  name   # :asset or :avatar
end
```

That means you can achieve a dynamic maximum size like this:

```rb
class Attachment < Shrine
  ...
  Attacher.validate do
    max_size = case record
               when Image then 5.megabytes
               when User  then 1.megabyte
               end

    validate_max_size(max_size)
  end
  ...
end
```

Considering that the model attachment definition is fully defined by the model and the attachment name, there isn't any need to pass in further options at that level. Also, there is already [`Shrine::Attacher#context`](http://shrinerb.com/rdoc/files/doc/attacher_md.html#label-Context) which can be modified when model is instantiated and also accessed from `Attacher.validate`, so `#options` would be redundant since it's basically `#context`.

PR #190, opened 2017-10-03
Fix copyright 2015-2016 => 2015-2017

yhirano55:
It seems that the end of period of copyright is 2017.
So I fixed it :)


janko-m:
Thanks!

PR #192, opened 2017-10-04
Update README.md Quickstart with commands

alexsydney:
* Added the filename and path for the initializer as a comment
* Added commands to generate the models and migration

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,7 @@ gem "shrine"
 ```
 
 ```rb
+# config/initializers/shrine.rb
 require "shrine"
 require "shrine/storage/file_system"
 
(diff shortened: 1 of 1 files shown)


janko-m:
Thanks for the patch, but I don't want to make the guide Rails-specific. In other web frameworks the configuration won't go in a `config/initializers/shrine.rb`, and the migration won't be generated with `rails g migration` and model with `rails g model`.

I would like to keep the documentation web framework agnostic, with specific Rails examples when it's useful. Readers can then decide where these pieces of code should go depending on the web framework they're using.

ltfschoen:
That's a shame, I'd have vouched this PR (or minor tweaks thereof) would have added value and avoided confused for numerous Rails users trying the gem

janko-m:
@ltfschoen I thought a bit more about this. The reason why I don't want to have Rails setup in the README is because I don't want users to think that this gem is Rails-specific (unlike Paperclip or CarrierWave).

However, I think a `shrine-rails` gem would be great, and its README would then include Rails-specific setup. The `shrine-rails` gem could contain generators like `rails generate shrine:install` `rails generate uploader`, and an `attachment_field` helper – like Refile does.

PR #199, opened 2017-10-18
Add frozen_string_literal

hmistry:


lib/shrine.rb
```diff
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
```
janko-m: I didn't know that the underscored `frozen_string_literal` works, but I would prefer if we used the dasherized `frozen-string-literal`, as that's officially the name of the feature.

hmistry: Interesting, I've always known it to be `frozen_string_literal` from all the articles/examples written about it and ruby style guides. However you're right in that the actual feature on Ruby code repo calls it using '-' instead of '_' but that's the only place I see from a quick search and I think it's the feature name that's related more with the command line option than the magic comment.

According to this [article](https://wyeworks.com/blog/2015/12/1/immutable-strings-in-ruby-2-dot-3), it's underscore but that's not official so take it as an example. There's Matsuda's [comment here](https://bugs.ruby-lang.org/issues/8976#note-30) which says underscore was chosen at the meeting with Matz, and here were the other [candidates (last page)](https://docs.google.com/document/d/1e00tTj8ix2ofS8H2RiIMUhr39bABSc7AL0vk4KbtSF4/edit) - all have underscores for the magic comment.

I can make your requested change - it's not a problem - though personally, I think the underscores follows ruby's conventional use of snake-case in naming methods and variables. Both work.

Would you still like me to change it?

lib/shrine/plugins/signature.rb
```diff
@@ -102,14 +104,14 @@ def calculate_md5(io)
...
-          crc = Zlib.crc32(io.read(16*1024, buffer ||= ""), crc) until io.eof?
+          crc = Zlib.crc32(io.read(16*1024, buffer ||= "".dup), crc) until io.eof?
           crc.to_s
         end
 
         def calculate_digest(name, io)
           require "digest"
           digest = Digest.const_get(name).new
-          digest.update(io.read(16*1024, buffer ||= "")) until io.eof?
+          digest.update(io.read(16*1024, buffer ||= "".dup)) until io.eof?
```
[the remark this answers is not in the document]
hmistry: Will do

lib/shrine.rb
janko-m: That's interesting, I didn't know that the underscored version was so wide-spread. I saw the dasherized version from Jeremy Evans' projects, and also in the `ruby --help` page. But I'm fine with the underscored version, let's keep it 👍

hmistry: Ok. Different worlds and I got to learn some Ruby history! 😃

PR #204, opened 2017-11-02
Delete tempfile if error occurs during download.

hmistry:
This addresses the issue of ensuring tempfiles are cleared in the event an error occurs during download in base, S3, and FileSystem modules. Implementation is straightforward like we discussed, the challenge was in testing.

### DIFF
--- a/lib/shrine.rb
+++ b/lib/shrine.rb
@@ -803,9 +803,9 @@ class Shrine
               tempfile = Tempfile.new(["shrine", ".#{extension}"], binmode: true)
               open(*args) { |io| IO.copy_stream(io, tempfile.path) }
               tempfile.tap(&:open)
-            rescue => err
+            rescue
               tempfile.close!
-              raise err
+              raise
             end
           end
         end
--- a/lib/shrine.rb
+++ b/lib/shrine.rb
@@ -799,9 +799,14 @@ class Shrine
           if storage.respond_to?(:download)
             storage.download(id, *args)
           else
-            tempfile = Tempfile.new(["shrine", ".#{extension}"], binmode: true)
-            open(*args) { |io| IO.copy_stream(io, tempfile.path) }
-            tempfile.tap(&:open)
+            begin
+              tempfile = Tempfile.new(["shrine", ".#{extension}"], binmode: true)
+              open(*args) { |io| IO.copy_stream(io, tempfile.path) }
+              tempfile.tap(&:open)
+            rescue => err
+              tempfile.close!
+              raise err
+            end
           end
         end
 
--- a/lib/shrine.rb
+++ b/lib/shrine.rb
@@ -804,7 +804,7 @@ class Shrine
               open(*args) { |io| IO.copy_stream(io, tempfile.path) }
               tempfile.tap(&:open)
             rescue
-              tempfile.close!
+              tempfile.close! if tempfile
               raise
             end
           end
--- a/lib/shrine/storage/file_system.rb
+++ b/lib/shrine/storage/file_system.rb
@@ -208,9 +208,9 @@ class Shrine
             tempfile = Tempfile.new(["shrine-filesystem", File.extname(args[0])], binmode: true)
             open(*args) { |file| IO.copy_stream(file, tempfile) }
             tempfile.tap(&:open)
-          rescue => err
+          rescue
             tempfile.close!
-            raise err
+            raise
           end
         else
           super
--- a/lib/shrine/storage/file_system.rb
+++ b/lib/shrine/storage/file_system.rb
@@ -203,10 +203,15 @@ class Shrine
       # Catches the deprecated `#download` method.
       def method_missing(name, *args)
         if name == :download
-          Shrine.deprecation("Shrine::Storage::FileSystem#download is deprecated and will be removed in Shrine 3.")
-          tempfile = Tempfile.new(["shrine-filesystem", File.extname(args[0])], binmode: true)
-          open(*args) { |file| IO.copy_stream(file, tempfile) }
-          tempfile.tap(&:open)
+          begin
+            Shrine.deprecation("Shrine::Storage::FileSystem#download is deprecated and will be removed in Shrine 3.")
+            tempfile = Tempfile.new(["shrine-filesystem", File.extname(args[0])], binmode: true)
+            open(*args) { |file| IO.copy_stream(file, tempfile) }
+            tempfile.tap(&:open)
+          rescue => err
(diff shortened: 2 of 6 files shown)


lib/shrine.rb
```diff
@@ -799,9 +799,14 @@ def download(*args)
...
-            tempfile = Tempfile.new(["shrine", ".#{extension}"], binmode: true)
-            open(*args) { |io| IO.copy_stream(io, tempfile.path) }
-            tempfile.tap(&:open)
+            begin
+              tempfile = Tempfile.new(["shrine", ".#{extension}"], binmode: true)
+              open(*args) { |io| IO.copy_stream(io, tempfile.path) }
+              tempfile.tap(&:open)
+            rescue => err
+              tempfile.close!
+              raise err
```
janko-m: If we call `raise` without an argument, Ruby will automatically raise the last rescued exception, so we could omit the exception variable here:

```rb
begin
  # ...
rescue
  # ...
  raise
end
```

lib/shrine.rb
```diff
@@ -799,9 +799,14 @@ def download(*args)
...
           else
-            tempfile = Tempfile.new(["shrine", ".#{extension}"], binmode: true)
-            open(*args) { |io| IO.copy_stream(io, tempfile.path) }
-            tempfile.tap(&:open)
+            begin
+              tempfile = Tempfile.new(["shrine", ".#{extension}"], binmode: true)
+              open(*args) { |io| IO.copy_stream(io, tempfile.path) }
+              tempfile.tap(&:open)
+            rescue => err
+              tempfile.close!
```
janko-m: If an exception occurs in `Tempfile.new` (e.g. there in case there are too many open file descriptors, so Ruby cannot create more), then the `tempfile` variable will be nil. We should handle that by closing the `tempfile` only if it was actually created:

```rb
tempfile.close! if tempfile
```

We should also test this scenario.

lib/shrine/storage/file_system.rb
```diff
@@ -203,10 +203,15 @@ def path(id)
...
-          tempfile.tap(&:open)
+          begin
+            Shrine.deprecation("Shrine::Storage::FileSystem#download is deprecated and will be removed in Shrine 3.")
+            tempfile = Tempfile.new(["shrine-filesystem", File.extname(args[0])], binmode: true)
+            open(*args) { |file| IO.copy_stream(file, tempfile) }
+            tempfile.tap(&:open)
+          rescue => err
+            tempfile.close!
+            raise err
+          end
```
janko-m: Same here:

```rb
begin
  # ...
rescue
  tempfile.close! if tempfile
  raise
end
```

lib/shrine/storage/s3.rb
```diff
@@ -254,10 +254,15 @@ def upload(io, id, shrine_metadata: {}, **upload_options)
...
-        tempfile.tap(&:open)
+        begin
+          (object = object(id)).get(response_target: tempfile, **options)
+          tempfile.singleton_class.instance_eval { attr_accessor :content_type }
+          tempfile.content_type = object.content_type
+          tempfile.tap(&:open)
+        rescue => err
+          tempfile.close!
+          raise err
+        end
```
janko-m: Same here:

```rb
begin
  # ...
rescue
  tempfile.close! if tempfile
  raise
end
```

test/uploaded_file_test.rb
```diff
@@ -319,6 +319,23 @@ def open(id)
...
+      @uploader.storage.instance_eval { undef download }
+      uploaded_file = @uploader.upload(fakeio("file"))
+      Tempfile.stub(:new, mock) do
+        IO.stub(:copy_stream, io_error) do
+          assert_raises(IOError) { uploaded_file.download }
+        end
+      end
+      mock.verify
+    end
+
```
janko-m: Instead of creating isolated mocks, what do you think about stubbing with actual `Tempfile` objects? I generally prefer that because it avoids me to use method call expectations. I would also prefer if we stubbed `Storage#open` to return an exception instead of `IO.copy_stream`, so that tests communicate what we're actually guarding against; `IO.copy_stream` will fail only if the source object is not a real IO, so for me that's not a concern as then it will consistently fail, but `Storage#open` can fail with temporary network exceptions, and that's why I wanted to guard against.

So I was thinking something in the lines of:

```rb
@uploader.storage.instance_eval { undef download }
@uploader.storage.instance_eval { def open(id); raise SystemCallError, "error occurred"; end }
uploaded_file = @uploader.upload(fakeio)
Tempfile.stubs(:new, tempfile = Tempfile.new) do
  assert_raises(SystemCallError) { uploaded_file.download }
end
assert tempfile.closed?
assert_nil tempfile.path # when tempfile gets deleted its #path becomes nil
```

janko-m: You could simplify this even more by creating a `Shrine::UploadedFile` object that doesn't exist (you can use the `#uploaded_file` helper method that already exists on top of the test file), and then `Shrine::Storage::Memory` (which is a superclass of the `Shrine::Storage::Test` we're working with here) will raise a `KeyError` because it uses `Hash#fetch` in `#open`.

```rb
@uploader.storage.instance_eval { undef download }
Tempfile.stubs(:new, tempfile = Tempfile.new) do
  assert_raises(KeyError) { uploaded_file.download }
end
assert tempfile.closed?
assert_nil tempfile.path # when tempfile gets deleted its #path becomes nil
```

lib/shrine.rb
hmistry: Good to know about the too many open files limitation. Agree.

lib/shrine.rb
hmistry: I did not know this and actually find it ambiguous. I like Ruby's tolerance for omitting things like parenthesis and stuff to reduce visual clutter in the code but I tend to draw the line when it makes the code ambiguous. If I saw this I'd think the code is raising raise's default exception and not the rescued exception. I'd have to consult the docs and/or know that Ruby is actually raising the rescued error.

That's my opinion however I'll go ahead and change it as that's subjective. If you change your mind, I can change it back too - not a problem.

test/uploaded_file_test.rb
hmistry: I like your approach of stubbing with actual `Tempfile` better. It also reduces having to create other methods in the stub so the code can run up to the target code of interest (e.g. having to add Tempfile#path).

I understand regarding `Storage#open` - makes sense. Will make the changes.

hmistry:
Made the requested edits. Thanks for your feedback. 😃

PR #205, opened 2017-11-02
Include CHANGELOG.md in the packaged gem

prognostikos:
As a heavy user of `bundle open` I really appreciate it if the CHANGELOG is part of the gem.


janko-m:
Agreed, thanks for the PR!

PR #207, opened 2017-11-03
Passing a block to Shrine::UploadedFile#open is now optional.

hmistry:
Please review this. I think this is what you wanted.
> It would be nice to be able to call Shrine::UploadedFile#open without a block, i.e. to make the block optional.

For the test I copied `#to_io` but I'm not sure why the object_id will be different for the same instance... could you explain what you are testing for here?
```ruby
# uploaded_file_test.rb - L241 and L377
assert_equal uploaded_file.object_id, uploaded_file.object_id
```

### DIFF
--- a/lib/shrine.rb
+++ b/lib/shrine.rb
@@ -788,11 +788,13 @@ class Shrine
         def open(*args)
           return to_io unless block_given?
 
-          @io = storage.open(id, *args)
-          yield @io
-        ensure
-          @io.close if @io
-          @io = nil
+          begin
+            @io = storage.open(id, *args)
+            yield @io
+          ensure
+            @io.close if @io
+            @io = nil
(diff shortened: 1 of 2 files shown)


janko-m:
Thanks, this is exactly what I had in mind!

Note that when you `return` in the method, the `ensure` block will still get executed. This means that the current code will return the closed IO object. So we should probably figure out a different strategy.

> For the test I copied #to_io but I'm not sure why the object_id will be different for the same instance... could you explain what you are testing for here?

Thanks for noticing that, it was accidental, the assertion should have been that `#to_io` memoizes the underlying IO object:

```rb
assert_equal uploaded_file.to_io.object_id, uploaded_file.to_io.object_id
```

I fixed this now in master.

PR #208, opened 2017-11-04
Refine aws-sdk deprecation message

mvasin:
When both 'aws-sdk-s3' and 'aws-sdk' are absent, this deprecation warning
```
SHRINE DEPRECATION WARNING: Using aws-sdk 2.x is deprecated and support for it will be removed in Shrine 3, use the new aws-sdk-s3 gem instead.
/Users/user/.rbenv/versions/2.4.2/lib/ruby/2.4.0/rubygems/core_ext/kernel_require.rb:55:in `require':LoadError: cannot load such file -- aws-sdk
```
is not ideal, as it claims that one uses `aws-sdk`, when he/she does not.

### DIFF
--- a/lib/shrine/storage/s3.rb
+++ b/lib/shrine/storage/s3.rb
@@ -7,8 +7,9 @@ begin
     raise "Shrine::Storage::S3 requires aws-sdk-s3 version 1.2.0 or above"
   end
 rescue LoadError
-  Shrine.deprecation("Using aws-sdk 2.x is deprecated and support for it will be removed in Shrine 3, use the new aws-sdk-s3 gem instead.")
-  require "aws-sdk"
+  if require "aws-sdk"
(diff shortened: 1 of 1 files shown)


janko-m:
Thanks for catching this! What do you think about simply swapping these two lines? That way we can avoid the conditional as the `require` will fail if aws-sdk is absent.

Note that `require` returns true/false to indicate whether something has already been required before, but it will raise an exception if the file/gem doesn’t exist.

janko-m:
Actually, it would still be good to suggest aws-sdk-s3 when both gems are absent, as you suggested.

mvasin:
It's a bit more involved then I thought. I see no options to elegantly solve this without potentially slowing down loading speed or cluttering the code with unnecessarily complicated constructions. So let's just ignore this; current deprecation message is not perfect, it but it gives some good guidance, that's enough.

PR #209, opened 2017-11-05
Improve error messages when both aws-sdk-s3 and aws-sdk gems are missing.

hmistry:
This is inspired by @mvasin solution (PR #208).


mvasin:
I also had this retries idea, but decided it's too messy for the quirk in question.

I like the way you use a proc. Overall, looks good.

janko-m:
Thanks for the pull request, but I think this solution is bit too complex. I managed to come up with a simpler solution that should work, I will push it shortly.

PR #210, opened 2017-11-05
Remove redundant jquery.iframe-transport.js

mvasin:
Dobra večer, Janko!

For educational proposes I decided to come up with [the slimmest app possible that will upload files directly to S3 with a progress bar](https://gist.github.com/mvasin/8af43b146814eafb32ed919fc017118b). Along the way I figured out that `jquery.iframe-transport.js` used in the demo is redundant; both my micro app and your demo works without it. Seems like you have cargo-culted it from somewhere else ;) Or maybe I miss something, that is even more likely.


janko-m:
> Dobra večer, Janko!

Dobra večer 😄 

> Along the way I figured out that `jquery.iframe-transport.js` used in the demo is redundant; both my micro app and your demo works without it. Seems like you have cargo-culted it from somewhere else ;)

You're absolutely right. I now came across [this wiki page](https://github.com/blueimp/jQuery-File-Upload/wiki/Browser-support#iframe-transport), and it seems that jQuery-File-Upload uses iframes only for old browsers. Thanks for removing it.

Do you think you could make the same PR for [shrine-rails-example](https://github.com/erikdahlstrand/shrine-rails-example)?

mvasin:
Yes, seems like it requires uploads via XHR, and the support is [pretty widespread](https://caniuse.com/#feat=xhr2).

Sure, I'll create a PR to rails example repo as well, sa zadovoljstvom!

PR #213, opened 2017-11-09
Add information about json/jsonb column type

aried3r:
It's only mentioned in the changelogs and readme, I thought it might be a good idea to promote this information a bit more.

I used monospace for `text`, `json` and `jsonb` since they declare data types, like in the [postgres docs](https://www.postgresql.org/docs/current/static/datatype-json.html).


janko-m:
@aried3r Good call, thanks for the patch!

PR #215, opened 2017-11-13
Clarify usage of signature plugin

### LINKED ISSUE #214: Documentation for signature plugin incomplete?
In [`signature.rb`](https://github.com/janko-m/shrine/blob/247f75ffd2cc27051c74337dda2da17c1027d35c/lib/shrine/plugins/signature.rb#L11-L16) you can read the following:

>The plugin adds a `calculate_signature` instance and class method to
>`Shrine`, which accepts the IO object and hashing algorithm and returns
>the calculated hash.
>
>     Shrine.calculate_signature(io, :md5)
>     #=> "9a0364b9e99bb480dd25e1f0284c8555"

Maybe I misinterpret things or am just nitpicking at the documentation, but I think the documentation makes it seem that adding `plugin :signature` adds the class method to `Shrine`, but that only happens when you do `Shrine.plugin :signature` in an initializer or somewhere, *not* if you just add it to the uploader.

I am currently migrating from Carrierwave to Shrine and modifying the [`uploader_to_shrine_data`](https://github.com/janko-m/shrine/blob/master/doc/carrierwave.md#migrating-from-carrierwave) method to my needs and found out that I do not have access to `Shrine.calculate_signature` unless I add it to the global config.

I guess it makes sense to not have it in the global namespace if you only use it in a single uploader, so basically I'm just asking if it's ok if I amend the documentation?

aried3r:
I'm not yet super happy with the documentation, but it's a start. Maybe you have a few comments?

Fixes #214

### DIFF
--- a/lib/shrine/plugins/signature.rb
+++ b/lib/shrine/plugins/signature.rb
@@ -4,9 +4,9 @@ class Shrine
   module Plugins
     # The `signature` plugin provides the ability to calculate a hash from file
     # content. The hash can then be used as a checksum or just as a unique
-    # signature of the file.
+    # signature of the file. It is recommended to load this plugin globally.
     #
-    #     plugin :signature
+    #     Shrine.plugin :signature
     #
     # The plugin adds a `calculate_signature` instance and class method to
     # `Shrine`, which accepts the IO object and hashing algorithm and returns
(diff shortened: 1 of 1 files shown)


janko-m:
Thanks for the patch. I hope you don't mind that I pushed my own tweak of it, as I felt like it was a bit redundant to talk about loading the plugin globally or locally, since the README already has the documentation for it.

aried3r:
I don't mind, awesome!

PR #219, opened 2017-11-16
Fix encode_content_disposition - replace all "+" with " ", not just the first one

ndbroadbent:
`encode_content_disposition` was using `sub` instead of `gsub`, so only the first `+` was being replaced.

I had a test PDF called "Demo PDF - Google Docs". When I opened the S3 URL in Chrome and pressed CMD+s to save the PDF, the filename was set to: "Demo PDF+-+Google+Docs". This PR fixes the issue, by replacing all of the `+` characters with spaces.

### DIFF
--- a/test/s3_test.rb
+++ b/test/s3_test.rb
@@ -184,8 +184,8 @@ describe Shrine::Storage::S3 do
       @s3.client.stub_responses(:put_object, -> (context) {
         @s3.client.stub_responses(:get_object, content_disposition: context.params[:content_disposition])
       })
-      @s3.upload(fakeio, "foo", content_disposition: 'inline; filename=""été bar.pdf""')
-      assert_equal "inline; filename=\"\"été bar.pdf\"\"", CGI.unescape(@s3.object("foo").get.content_disposition)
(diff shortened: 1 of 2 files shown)


test/s3_test.rb
```diff
@@ -184,8 +184,8 @@ def filesystem
       @s3.client.stub_responses(:put_object, -> (context) {
         @s3.client.stub_responses(:get_object, content_disposition: context.params[:content_disposition])
       })
-      @s3.upload(fakeio, "foo", content_disposition: 'inline; filename=""été bar.pdf""')
-      assert_equal "inline; filename=\"\"été bar.pdf\"\"", CGI.unescape(@s3.object("foo").get.content_disposition)
+      @s3.upload(fakeio, "foo", content_disposition: 'inline; filename=""été foo bar.pdf""')
+      assert_equal "inline; filename=\"\"été foo bar.pdf\"\"", CGI.unescape(@s3.object("foo").get.content_disposition)
```
ndbroadbent: This doesn't actually test the escaping behavior, but I just thought the filenames should be consistent.

janko-m:
@ndbroadbent Good catch, thank you for fixing this!

PR #223, opened 2017-11-23
Fix typos in upload_endpoint.rb

aried3r:
Not sure if you accept cosmetic PRs like this. But here it is :D


janko-m:
@aried3r Thank you, typo corrections are always welcome! 😉

PR #233, opened 2017-12-16
Add Github issue and PR templates

hmistry:
I think this can be good in guiding users fill out relevant information for issues and pr. Also might help direct more people to the google group for questions and leave issues for bugs. It's worth trying...

Feel free to modify the templates as you see fit. 😃


janko-m:
Thank you for the initiative! I would definitely like to add an issue template, though for now I don't think the pull request template is necessary (as I was always able to understand what the PR was about so far).

Just a minor question, it seems to me that the `Steps to reproduce` and `Code to demonstrate issue` sections ask for pretty much the same thing, do you think it would make sense to merge them into a single section? I like Sequel's `Simplest possible self-contained example showing the bug`, and in the placeholder we could link to the templates from `CONTRIBUTING.md`.

hmistry:
Oops I missed that, yes they (steps to reproduce and code to demo) are the same and should be merged. Good catch.

I like the Sequel phrasing however very few people have been following it or any instructions. 🤷‍♂️  Another thought I have is separating the templates from `Contributing.md` doc because it doesn't fit the user mental model. If I'm posting an issue or asking a question, I would never think I'm contributing to Shrine gem and as such will never look at that doc. I think it would be nice to have the templates in a "playground" folder as a easy way for dev's to take a test drive of Shrine or debug or explore.

Let me take another pass on the issue template. I'll toss the PR template.

janko-m:
@hmistry Looks good, thank you!

PR #247, opened 2018-02-05
Fixed bug in logging plugin so subclasses correctly inherits configuration

### LINKED ISSUE #246: Shrine.logger.level value is not herited
A configuration problem similar to the now fixed https://github.com/janko-m/shrine/issues/57 .

Documentation says:
> Logging is by default disabled in tests, but you can enable it by setting Shrine.logger.level = Logger::INFO.

Based on this, I tried to change log level:

```ruby
Shrine.plugin :logging

class SomeUploader < Shrine; end

Shrine.logger.level = Logger::WARN
#=> 2
SomeUploader.logger.level
#=> 1
```

What happen here if I understand correctly, is that when the `:logger` option is not specified, each uploader class end up with a different `@logger` instance variable (cf. [here](https://github.com/janko-m/shrine/blob/master/lib/shrine/plugins/logging.rb#L66) and [here](https://github.com/janko-m/shrine/blob/master/lib/shrine/plugins/logging.rb#L122) ).

I expected `Shrine.logger.level` to be shared accross all uploaders. But as is, it needs to be set for each class.

I'm willing to make a pull request, but I'm not sure of the best solution:
- change `@logger` object to be uniq accross all uploaders ?
- keep one logger by class, but add an option to choose logger level at plugin initialisation ?

Thanks for your time, this gem is awesome.

hmistry:
Fixed a bug in the logging plugin where the logger of Shrine's subclass does not inherit the logging level from Shrine. With this fix, the subclass now correctly inherits the logging level from Shrine and is a copy of the logger so it can be independently configured if needed without affecting the base Shrine logger or other subclass's logger configuration.

I left the default logging level to Logger::INFO which is the current default previously set in the code. This is different from what's stated in issue #246.

Tests added to ensure the behavior reflects the correct inheritance model as described in the issue.

Fixes #246.

### DIFF
--- a/lib/shrine/plugins/logging.rb
+++ b/lib/shrine/plugins/logging.rb
@@ -60,7 +60,7 @@ class Shrine
         # Assigns the subclass a copy of the logger.
         def inherited(subclass)
           super
-          subclass.opts[:logging_logger] = subclass.opts[:logging_logger].dup
+          subclass.opts[:logging_logger] = subclass.opts[:logging_logger].dup if subclass.opts[:logging_logger]
         end
 
         def logger=(logger)
--- a/lib/shrine/plugins/logging.rb
+++ b/lib/shrine/plugins/logging.rb
@@ -51,32 +51,26 @@ class Shrine
       end
 
       def self.configure(uploader, opts = {})
-        uploader.opts[:logging_logger] = opts.fetch(:logger, uploader.opts[:logging_logger])
         uploader.opts[:logging_stream] = opts.fetch(:stream, uploader.opts.fetch(:logging_stream, $stdout))
+        uploader.opts[:logging_logger] = opts.fetch(:logger, uploader.opts.fetch(:logging_logger, uploader.create_logger))
         uploader.opts[:logging_format] = opts.fetch(:format, uploader.opts.fetch(:logging_format, :human))
       end
 
       module ClassMethods
-        # Assigns the subclass a copy of the logger.
-        def inherited(subclass)
-          super
-          subclass.opts[:logging_logger] = subclass.opts[:logging_logger].dup if subclass.opts[:logging_logger]
-        end
-
         def logger=(logger)
-          @logger = logger || Logger.new(nil)
+          @logger = logger
         end
 
-        # Initializes a new logger if it hasn't been initialized.
         def logger
-          @logger ||= opts[:logging_logger] || (
-            logger = Logger.new(opts[:logging_stream])
-            logger.level = Logger::INFO
-            logger.level = Logger::WARN if ENV["RACK_ENV"] == "test"
-            logger.formatter = pretty_formatter
-            opts[:logging_logger] = logger
-            logger
-          )
+          @logger ||= opts[:logging_logger]
+        end
+
+        def create_logger
+          logger = Logger.new(opts[:logging_stream])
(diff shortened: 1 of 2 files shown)


lib/shrine/plugins/logging.rb
```diff
@@ -68,6 +74,7 @@ def logger
             logger.level = Logger::INFO
             logger.level = Logger::WARN if ENV["RACK_ENV"] == "test"
             logger.formatter = pretty_formatter
+            opts[:logging_logger] = logger
```
janko-m: I would prefer if `Shrine.opts` is modified only when loading plugins (in the `configure` method), I think that modifying it dynamically like that when invoking reader methods could potentially lead so subtle bugs.

I was thinking that we could move all this code for creating the default logger into the `configure` method.

lib/shrine/plugins/logging.rb
```diff
@@ -57,6 +57,12 @@ def self.configure(uploader, opts = {})
       end
 
       module ClassMethods
+        # Assigns the subclass a copy of the logger.
+        def inherited(subclass)
+          super
+          subclass.opts[:logging_logger] = subclass.opts[:logging_logger].dup if subclass.opts[:logging_logger]
+        end
```
janko-m: After thinking about it more, I think it actually isn't a bad idea to share the `Logger` object between uploaders, in which case we could remove this code. I feel like that then we'd need to do the same in other plugins for consistency, and I don't see the need really. What do you think?

hmistry: I think a Logger instance should be shared in an app because I'd like to control logging level globally throughout the app. Then individual messages are controlled by using message level. Duplicating the logger in subclassing prevents the propagation of the global logger level to subclasses if done after subclassing. However I think it's ok to duplicate logger instance in multi-process/thread applications to maintain isolation between them.

What was your reasoning for duplicating logger in subclass? Are there any particular use case that calls for it?

As far as making other plugins consistent, I presume you mean the duplicating of their respective options or something else? I think it's nice to have consistency but that's a bigger change and we should review each plugin to see if it makes sense or not... something better done for next major release.

lib/shrine/plugins/logging.rb
hmistry: Agree. I'll make the changes.

lib/shrine/plugins/logging.rb
janko-m: > What was your reasoning for duplicating logger in subclass? Are there any particular use case that calls for it?

No, I thought it made sense, but it doesn't anymore.

> As far as making other plugins consistent, I presume you mean the duplicating of their respective options or something else? I think it's nice to have consistency but that's a bigger change and we should review each plugin to see if it makes sense or not... something better done for next major release.

Yeah, what I meant to say is that I'm in favour of *not* duplicating the `Logger` instance, but having subclasses use the same `Logger` instance. Because if we were to duplicate the `Logger` instance, that would not be consistent with any other potential plugin that's not duplicating it's option values. So I don't think it makes sense to duplicate the `Logger` instance at this point.

hmistry: Ok great so we're in agreement for not duplicating logger. I'll go ahead and make this change.

janko-m:
@hmistry Perfect, thanks a lot! ❤️

PR #255, opened 2018-02-23
Prefer registered MIME types with the mime-types analyzer

aprescott:
(This is potentially a breaking change, depending on how strict downstream applications are about media type return values.)

This was specifically written to make CSV file extensions map to text/csv instead of text/comma-separated-values when using the mime-types gem, but will more generally prefer registered types over un-registered types.

The IANA Media Types registry (referencing RFC 4180 and RFC 7111) say that the media type for CSV is text/csv. Because of the way the mime-types gem implements MIME::Types.of(filename), you get text/comma-separated-values as listed first for a CSV:

```ruby
MIME::Types.of("csv")
# => [#<MIME::Type: text/comma-separated-values>, #<MIME::Type: text/csv>]
```

Since `MIME::Type#registered?` indicates that `text/csv` is a registered type, this change makes that the preferred media type for CSV files.

See mime-types/ruby-mime-types#92 for a related issue about fixing this in the mime-types gem.

### DIFF
--- a/lib/shrine/plugins/determine_mime_type.rb
+++ b/lib/shrine/plugins/determine_mime_type.rb
@@ -201,7 +201,7 @@ class Shrine
           require "mime/types"
 
           if filename = extract_filename(io)
-            mime_type = MIME::Types.of(filename).first
+            mime_type = MIME::Types.of(filename).partition(&:registered?).flatten.first
             mime_type.content_type if mime_type
           end
         end
--- a/test/plugin/determine_mime_type_test.rb
+++ b/test/plugin/determine_mime_type_test.rb
@@ -121,6 +121,10 @@ describe Shrine::Plugins::DetermineMimeType do
     it "returns nil for empty IOs" do
       assert_nil @shrine.determine_mime_type(fakeio(""))
     end
+
+    it "prioritizes text/csv as registered, over text/comma-separated-values" do
+      assert_equal "text/csv", @shrine.determine_mime_type(fakeio(filename: "report.csv"))
+    end
   end
 
   describe ":mini_mime analyzer" do


aprescott:
Some thoughts/concerns here:

1. This is maybe a breaking change.
2. The test I added is specific to `text/csv` and `text/comma-separated-values`, but doesn't test against `registered?` return values directly. I didn't know whether it'd be better to change the testing approach for one method to start stubbing `Types.of()`.
3. This arguably should be fixed upstream in `of()`, but there didn't seem to be a lot of traction, based on https://github.com/mime-types/ruby-mime-types/pull/92, so I thought I'd PR this anyway.

For the application I'm currently working on, I only have this issue with `text/csv`, so I had a very clear work-around in the plugin configuration:

```ruby
if mime_type == "text/plain"
  mime_type = analyzers[:mime_types].call(io)
end

if mime_type == "text/comma-separated-values"
  mime_type = "text/csv"
end
```

To fix this generally, though, I think I'd have to monkey-patch or start fiddling with the mime-types configuration. Since I think this is a bug, I'd like to avoid all my workarounds, if possible!

janko-m:
@aprescott Thanks a lot for the PR!

I ran the following script to see which MIME types change, and it doesn't turn out to be many:

```rb
MIME::Types.map(&:preferred_extension).compact.uniq.each do |extension|
  content_type1 = MIME::Types.of(extension).first.content_type
  content_type2 = MIME::Types.of(extension).partition(&:registered?).flatten.first.content_type
  puts "Extension: .#{extension}, content_type1: #{content_type1.inspect}, content_type2: #{content_type2.inspect}" if content_type1 != content_type2
end
```
```
Extension: .xls, content_type1: "application/excel", content_type2: "application/vnd.ms-excel"
Extension: .class, content_type1: "application/java-vm", content_type2: "application/octet-stream"
Extension: .wks, content_type1: "application/lotus-123", content_type2: "application/vnd.lotus-1-2-3"
Extension: .mcd, content_type1: "application/mathcad", content_type2: "application/vnd.mcd"
Extension: .ppt, content_type1: "application/powerpoint", content_type2: "application/vnd.ms-powerpoint"
Extension: .hlp, content_type1: "application/winhlp", content_type2: "text/plain"
Extension: .wp, content_type1: "application/wordperfect", content_type2: "application/wordperfect5.1"
Extension: .t, content_type1: "application/x-troff", content_type2: "text/troff"
Extension: .man, content_type1: "application/x-troff-man", content_type2: "text/troff"
Extension: .me, content_type1: "application/x-troff-me", content_type2: "text/troff"
Extension: .ms, content_type1: "application/x-troff-ms", content_type2: "text/troff"
Extension: .dwf, content_type1: "drawing/dwf", content_type2: "model/vnd.dwf"
Extension: .csv, content_type1: "text/comma-separated-values", content_type2: "text/csv"
```

However, it would still be a backwards incompatible change for users accepting uploads for these types. I really think this should be fixed in the `mime-types` gem itself, I would prefer to avoid patching other gems in Shrine if it can be avoided. I expect the reason the PR didn't gain much traction is because the author closed the PR. I just opened a new one: https://github.com/mime-types/ruby-mime-types/issues/131.

janko-m:
I'm closing this in favour of https://github.com/mime-types/ruby-mime-types/pull/132.

PR #260, opened 2018-03-15
Remove broken article link from homepage

mokolabs:
This link is dead.

The site is offline, the domain has expired, and the content has not been reposted elsewhere. 🙁


janko-m:
That's a shame. Good catch, thanks!

mokolabs:
Yeah, I was excited to read that one myself. Oh well.

PR #262, opened 2018-03-16
Sort metadata for download_identifier

### LINKED ISSUE #261: URLs generated by download_endpoint are not always staying the same for same file
### Brief Description
URLs generated by `download_endpoint` not always the same for same file.

### Expected behavior
Generated URL should always be the same for the same file.

### Actual behavior
URL generation takes metadata fields in random order. Or at least not fixed order. So under certain circumstances the base64 string changes. E.g. I'm seeing this when generating a URL from a model that just assigned a file vs. a model loaded from database.

https://github.com/shrinerb/shrine/blob/4dc52969b06c6287496be5a053afadd81b1e66a4/lib/shrine/plugins/download_endpoint.rb#L133

### System configuration
**Ruby version**: 2.4.3p205

**Shrine version**: 2.9.0

MSchmidt:
Sorting the metadata values by key, so the download_identifier always looks the same regardless of the order of metadata fields. Fixes #261.

### DIFF
--- a/lib/shrine/plugins/download_endpoint.rb
+++ b/lib/shrine/plugins/download_endpoint.rb
@@ -131,7 +131,7 @@ class Shrine
         # long.
         def download_identifier
           semantical_metadata = metadata.select { |name, _| %w[filename size mime_type].include?(name) }
-          download_serializer.dump(data.merge("metadata" => semantical_metadata))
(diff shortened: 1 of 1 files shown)


hmistry:
Thank you for the PR. Could you add a test to check for it?

janko-m:
@MSchmidt Thanks for the PR!

PR #267, opened 2018-04-04
Update external.yml

edwardsharp:
i forked and fixed up the existing Google Drive storage gem (verynear/shrine-google_drive_storage) to use a Google API Service Account. 

it's passing the shrine lint tests!


www/_data/external.yml
```diff
@@ -131,7 +131,7 @@
         url: https://github.com/renchap/shrine-google_cloud_storage
       -
         name: shrine-google_drive_storage
```
janko-m: We should also update the name

PR #270, opened 2018-04-13
Fix typo in :infer_extension plugin docs

mokolabs:


janko-m:
Thanks for the patch!

PR #277, opened 2018-06-21
Update Active Record dependency to >= 5.2

hmistry:


shrine.gemspec
```diff
@@ -51,10 +51,15 @@ direct uploads for fully asynchronous user experience.
   end
 
   gem.add_development_dependency "sequel"
-  gem.add_development_dependency "activerecord", "~> 4.2"
+
+  if RUBY_VERSION >= "2.3.0"
```
janko-m: Rails 5.x versions require MRI 2.2.2+, so maybe we could update the conditional to be `if RUBY_VERSION >= 2.2.2`?

In that case we can even loosen the conditional on ActiveRecord to be `~> 5.0`.

What do you think?

test/plugin/activerecord_test.rb
```diff
@@ -146,7 +146,11 @@
     it "isn't triggered when callback chain is halted" do
       @user.update(avatar: fakeio)
       uploaded_file = @user.avatar
-      @user.class.before_save { false }
+      if RUBY_VERSION >= "2.3.0"
```
janko-m: Here as well, I would check on the ActiveRecord version.

shrine.gemspec
```diff
@@ -51,10 +51,15 @@ direct uploads for fully asynchronous user experience.
...
+    gem.add_development_dependency "activerecord", ">= 5.2"
+  else
+    gem.add_development_dependency "activerecord", "~> 4.2"
+  end
 
   if RUBY_ENGINE == "jruby"
-    gem.add_development_dependency "activerecord-jdbcsqlite3-adapter", "1.3.24"
+    gem.add_development_dependency "activerecord-jdbcsqlite3-adapter"
```
janko-m: It seems that Travis CI failures are related to the fact that `activerecord-jdbcsqlite3-adapter` 50.0 was installed, and that one isn't compatible with Rails 5.2. That adapter works in a way that version 50.x is compatible with Rails 5.0, version 51.x with Rails 5.1, version 52.x with Rails 5.2 and so on.

It seems that version 52.0 hasn't been released yet, the latest one is 51.1 (I don't understand why that one wasn't installed). I think on JRuby we might need to constrain ActiveRecord to `< 5.2` to make the CI pass. Maybe on JRuby we should just specify `gem "activerecord", "5.1"` and `gem "activerecord-jdbcsqlite3-adapter", "51"` to make sure compatible versions are used.

hmistry:
@janko-m Thanks a lot for your feedback!

Issue #276, opened 2018-06-17
Improve docs for multiple uploads


devkir:
- Issues are for bug reports only.
- Have a question or unsure if it's a bug, then please inquire on the [ruby-shrine Google Group](https://groups.google.com/forum/#!forum/ruby-shrine).
---

### Brief Description
Please update the docs for multiple uploads with Shrine and ActiveRecord. Currently there is no information what params should controller accept. How to validate presence of has_many :images on parent model. 

### Expected behavior
Better docs

### System configuration
**Ruby version**: 2.5.1

**Shrine version**: 2.1.1

janko-m:
Note that the guide assumes you already know how ActiveRecord's [nested attributes](http://api.rubyonrails.org/classes/ActiveRecord/NestedAttributes/ClassMethods.html) work. But fair enough, I'll try to make the docs clearer.

> Currently there is no information what params should controller accept.

From the guide, these are the parameters that should be sent to the controller (for a `Album.has_many :photos` scenario):

```rb
album[photos_attributes][0][image] = '{"id":"38k25.jpg","storage":"cache","metadata":{...}}'
album[photos_attributes][1][image] = '{"id":"sg0fg.jpg","storage":"cache","metadata":{...}}'
album[photos_attributes][2][image] = '{"id":"041jd.jpg","storage":"cache","metadata":{...}}'
# ...
```

Then in your controller you would just pass these parameters when creating/updating the Album:

```rb
Album.create(params[:album])
```

The `:photos_attributes` attribute is generated from `Album.accepts_nested_attributes_for :photos`, and works as described in the ActiveRecord documentation. It doesn't matter whether the `image` is a Shrine attachment attribute or just a regular column attribute, it's all the same.

> How to validate presence of has_many :images on parent model.

See the ActiveRecord documentation on how to validate presence of associated records, this is not related to Shrine.

devkir:
I tried on my controller `params.require(:parent_model).permit(photos_attributes: [:id, :image]` - it didn't work, validation also didn't worked. 
_Maybe it is because Photo model is polymorphic?_

janko-m:
There could be a number of reasons why this didn't work. My guess is that you have to use `images_attributes` attribute, not `photos_attributes`, as you said you have `has_many :images`.

I would suggest you try to isolate your example in a script to prove that it works, and then debug why it doesn't work in your app. Here is my example proving that it works:

```rb
require "active_record"
require "shrine"
require "shrine/storage/file_system"

[...]

PR #281, opened 2018-06-30
Improve multiple file uploads doc

hmistry:
Addresses #276

### DIFF
--- a/doc/multiple_files.md
+++ b/doc/multiple_files.md
@@ -1,17 +1,11 @@
 # Multiple Files
 
-There are often times when you want to allow users to attach multiple files to
-a single resource. Some file attachment libraries provide a special interface
-for multiple attachments, but Shrine doesn't come with one, because it's much
-more robust and flexible to implement this using your ORM directly.
+There are times when you want to allow users to attach multiple files to a single resource like an album having many photos or a playlist having many songs. Some file attachment libraries provide a special interface for multiple attachments, but Shrine doesn't have one because it's much more robust and flexible to implement this using your ORM directly.
 
-The idea is to create a new table, and attach each uploaded file to a separate
-record on that table, while having a "many-to-one" relationship with the main
-table. That way a database record from the main table can implicitly have
-multiple attachments through the associated records.
+The basic idea to implement this in Shrine is to create a separate files table from the main resource table, and attach each uploaded file to a separate record on the files table. Next you'll use a "one-to-many" relationship between the main resource table and the files table to track which files belongs to which main resource just like you do with other "one-to-many" associated database models. Each database record from the main table can implicitly have multiple attachments through the associated records.
 
 ```
-album
+album1
   photo1
     - attachment1
   photo2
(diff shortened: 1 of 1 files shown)


janko-m:
Sorry it took so long to merge, I wanted to add many more details to the doc. I kept as much of your changes as possible, they helped a lot. Just a few comments:

* markdown documents should remain hard-wrapped (at least according to my preference)
* `rdoc` doesn't take kindly to bullets not being separated from the paragraph in markdown documents, so this will be correctly interpreted

  ```
  This is a paragraph.

  * these
  * are
  * bullets
  ```

  but this won't

  ```
  This is a paragraph.
  * these
  * are
  * bullets
  ```

hmistry:
I like your improvements a lot. ❤️ 

The key was finding a format (i.e. grouping into steps) that best guided the reader whilst organizing the information. I felt the view part was weak but wasn't sure how to work around the imposed documenting view limitations. Now it's a lot more complete with your form and json examples.

marciodat:
Hi there! 

Despite of the huge improvement on the documentation I would like to make a suggestion. To be honest I'm still confused on how to set up this using just the information provide on the help alone (I've been struggling for almost a week now and still haven't figure out...). 

Maybe you could add some more information on how to deal with this: 

`file_field_tag "files[]", multiple: true`

on the controller. It will be very useful. I tried this: https://stackoverflow.com/a/41846249/6706893 and It didn't work. 

Code on stackoverflow:

```
class ProductsController < ApplicationController
  def create
    images_attributes = params["files"].map { |file| {image: file} }
    Product.create(product_params.merge(images_attributes: images_attributes))
  end
end
```

I don't really know what to do and I'm sure that there are probably a lot of people on the same situation.

Even looking on the demo app didn't help. It's also not clear there. 

Thanks!

hmistry:
Hi @marciodat 

Thank you for the feedback. We'll revisit the documentation and see how we can improve it.

If I understand correctly, the issue you're having has more to do with how Rails `accepts_nested_attributes` work between forms and controller and less to do with Shrine. We are using `accepts_nested_attributes` because we're trying to create a new album with multiple photos in one go. The form has to build the right data structure with all of it fields and multiple files and nesting correctly so the `params` hash in the Rails controller matches the Rails parent model and the child models relationship accordingly. With a valid `params`, Rails will know how to create/update the models.

If you clone the Rails demo, set a [breakpoint here](https://github.com/erikdahlstrand/shrine-rails-example/blob/master/app/controllers/albums_controller.rb#L11), and inspect `album_params` that the form sent to the controller. See how the hash data structure album with photo files matches the `Album` and has many related `Photo` models. In the `Album` model, we allow [nested attributes here](https://github.com/erikdahlstrand/shrine-rails-example/blob/master/app/models/album.rb#L3) so Rails can also create/update child photos. Compare with your app and see if your form is sending the correct `params` hash that matches your models so your Controller and Rails can build it. 

Here's [one reference](https://rubyplus.com/articles/3681-Complex-Forms-in-Rails-5) on Rails nested attributes. Hope it helps.

marciodat:
Thanks for the quick answer! 

I really like your gem. That's why I've been struggling to make it work for so long. I don't like the other options. 

The problem is that  I don't have something like this on my app:

```
  <ul class="list-unstyled file-upload-list" id="album-photo-list">
    <% @album.photos.each_with_index do |photo, idx| %>
      <%= render partial: "photo", locals: { photo: photo, idx: idx } %>
    <% end %>
  </ul>
```
The partial kind of "prepare" the params for the controller arranging then in the correct format. I think I will have to create something with javascript to have the proper format of the params in `files[]`. 

I was trying to deal with `files[]` on the controller but so far I could not make it work.  The files array is just and array with strings with the name of the files. I haven't figured out how to "translate" that to have something like it should be: 

```
 ... "photos_attributes"=>{"jm218sx2c41rmq997"=>{"image"=>"{\"id\":\"e89db64cdc1e558e05
a66aa56794adb6.jpg\",\"storage\":\"cache\",\"metadata\":{\"filename\":\"photos-1.jpg\",\"size\":7149,\"mime_type\":\"image/jpeg\"}}", "title"=>"some title"}}},
```

I'm dealing with unknown files. It could be a pdf, or maybe a ods, sometimes a image. I don't want to show a preview. Just the name of the file and a delete option. 

Maybe, if possible, you could add some tips on the help on how to deal with that (without using Uppy - if possible). If I figured out how to do it I can help you if you want.


Thanks for the amazing work!

hmistry:
@marciodat This PR is for documentation so we'd like to keep only relevant discussion around that.

If you want help on your issue, please post the issue details on our [Shrine Google Group](
https://groups.google.com/forum/#!forum/ruby-shrine).

Provide us with details on:
1. What's your problem and what's your goal?
2. Provide code samples for:
    1. relevant Shrine Uploader
    1. Model
    1. Controller
    1. View - can be the form or the request body its sending
3. Any other relevant information that we might need to know. 

Thanks.

janko-m:
I added the JavaScript-less approach in the "Multiple Files" guide: https://github.com/shrinerb/shrine/blob/051f72622d4dc2dbce2f42e8e7c4251cd14ce8ab/doc/multiple_files.md#4a-form-upload

Writing comprehensive generic docs which don't assume the web framework, ORM, or the client side file upload library can be really challenging sometimes, I hope this guide has enough information now 😃

PR #282, opened 2018-07-03
Add comments to Roda demo code and link to Rails demo

hmistry:
This is what I had in mind - little notes to guide users along. Let me know what you think.


hmistry:
I also tweaked the Wiki if you didn't see it. I thought it would go through a review but it made the changes live.

PR #285, opened 2018-07-19
Handle nil data_uri's in addition to blank

GeekOnCoffee:


janko-m:
Actually, I'm wondering whether this is a good idea, because now if the user is accidentally assigning a nil value for data URI (e.g. by making a typo when fetching from params), they will not get notified and may spend some time debugging. What is your use case that benefits from this change?

PR #294, opened 2018-09-18
Reduce multiple method calls in some methods

hmistry:


lib/shrine.rb
```diff
@@ -617,18 +620,21 @@ def url(**options)
...
+          file = get
+          file && store.uploaded?(get)
         end
 
         # Returns a Shrine::UploadedFile instantiated from the data written to
         # the attachment attribute.
         def get
-          uploaded_file(read) if read
+          value = read
+          uploaded_file(value) if value
```
janko-m: I don't think this change is necessary, as `read` literally just calls the cached record's column value, so there is no performance penalty here other than the overhead of a method call, which we don't need to optimize for.

hmistry:
@janko-m I'll make the changes. In this case, brevity made the code harder to read for me because reading `get && store.uploaded?(get)` is not as clear as reading:
```
file = get 
file && store.uploaded?(file)
```
Because `get` doesn't hint at what the object is. Similar for `read`.

FYI these changes resulted in a 10% overall reduction in method calls (which I know may not be computational intensive) when tracing through a single original file being attached to a record. It resulted in a 20% reduction when the versions plugin was added and the process block creates one original file and one version.

PR #299, opened 2018-09-29
[WIP] Add a concept version of new derivates plugin

hmistry:
This is v0.1 concept of the new derivates plugin. It's not complete in any form whatsoever. 
Submitting code to facilitate discussion of implementation direction and API development.

This concept can either store derivates data under the original db column "{name}_data" or a separate derivates db column called "{name}_derivates_data". To configure the `derivates` plugin, pass `data_attribute: :original` or `data_attribute: :separate` to the plugin.

Still ways to go. This can all change if needed. 😃 

Here is a test script.

```ruby
require 'bundler/setup'

require "active_record"
require "shrine"
require 'byebug'
require 'pp'
require '../test/support/fakeio'

  Shrine.storages = {
    cache: Shrine::Storage::FileSystem.new(Dir.tmpdir, prefix: "cache"),
    store: Shrine::Storage::FileSystem.new(Dir.tmpdir, prefix: "store"),
  }


  ActiveRecord::Base.establish_connection(adapter: "sqlite3", database: ":memory:")
  ActiveRecord::Base.connection.create_table(:photos) do |t|
    t.text :image_data
  end
  ActiveRecord::Base.connection.create_table(:videos) do |t|
    t.text :vid_data
    t.text :vid_derivates_data
  end


  # Uses current versions plugin
  class MyUploader < Shrine
    plugin :activerecord
    plugin :processing
    plugin :versions

      { original: io, small: io.download }
    end
  end

  class Photo < ActiveRecord::Base
    include MyUploader::Attachment.new(:image)
  end


  # Uses new derivates plugin
  class NewUploader < Shrine
    plugin :activerecord
    plugin :derivates, data_attribute: :original  # store under original file db column
    # plugin :derivates, data_attribute: :separate  # store under separate derivates db column
  end

  class Video < ActiveRecord::Base
    include NewUploader::Attachment.new(:vid)
  end

  file = FakeIO.new("Hello World", filename: "hw.txt")


puts "\n----- Versions ------"

  photo = Photo.create(image: file)
  pp JSON.parse(photo.image_data || "{}")


puts "\n----- Derivates ------"

  file = FakeIO.new("Hello World", filename: "hw.txt")

  video = Video.new
  video.vid = file
  video.save

  pp JSON.parse(video.vid_data || "{}")
  pp JSON.parse(video.vid_derivates_data || "{}")

  sfile = FakeIO.new("Hello Smallville", filename: "shw.txt")
  video.vid_attacher.add_derivate(:small, sfile)

  mfile = FakeIO.new("Hello Metropolis", filename: "mhw.txt")
  video.vid_attacher.add_derivate(:medium, mfile)
  video.vid_attacher.update_derivates
  video.save

  pp JSON.parse(video.vid_data || "{}")
  pp JSON.parse(video.vid_derivates_data || "{}")

puts "----- Done ------\n"
```

### DIFF
--- a/lib/shrine/plugins/derivates.rb
+++ b/lib/shrine/plugins/derivates.rb
@@ -1,0 +1,86 @@
+# frozen_string_literal: true
+
+class Shrine
+  module Plugins
+    module Derivates
+      def self.load_dependencies(uploader, *)
+        uploader.plugin :default_url
+      end
+
+      def self.configure(uploader, opts = {})
+        uploader.opts[:version_fallbacks] = opts.fetch(:fallbacks, uploader.opts.fetch(:version_fallbacks, {}))
+        uploader.opts[:versions_fallback_to_original] = opts.fetch(:fallback_to_original, uploader.opts.fetch(:versions_fallback_to_original, true))
+      end
+
+      module ClassMethods
+        def version_fallbacks
+          opts[:version_fallbacks]
+        end
+      end
+
+
+      module AttacherMethods
+        def changed?
+          instance_variable_defined?(:@old) || instance_variable_defined?(:@derivates_queue)
+        end
+
+        def finalize
+          super
+          update_derivates
+        end
+
+        def add_derivate(key, io, store = nil, nested_list = nil)
+          @derivates_queue ||= {}
+
+          @derivates_queue.merge! key.to_sym => store!(io)
+        end
+
+        def update_derivates
+          if instance_variable_defined?(:@derivates_queue)
+            _set_derivates(@derivates_queue)
+            remove_instance_variable(:@derivates_queue)
+          end
+        end
+
+        private
+
+        def _set_derivates(uploaded_file)
+          data = convert_to_data(uploaded_file) if uploaded_file
+          write_derivates(data ? convert_before_write(data) : nil)
+        end
+
+        def derivates_data_attribute
+          :"#{name}_derivates_data"
+        end
+
+        def read_derivates
+          value = record.send(derivates_data_attribute)
+          convert_after_read(value) unless value.nil? || value.empty?
+        end
+
+        def write_derivates(value)
+          record.send(:"#{derivates_data_attribute}=", value)
+        end
+
+        def fallback_to_original?
+          shrine_class.opts[:versions_fallback_to_original]
+        end
+
+        # Converts the Hash/Array of UploadedFile objects into a Hash/Array of data.
+        def convert_to_data(object)
+          if object.is_a?(Hash)
+            object.inject({}) do |hash, (name, value)|
+              hash.merge!(name => convert_to_data(value))
+            end
+          elsif object.is_a?(Array)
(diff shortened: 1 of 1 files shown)


lib/shrine/plugins/derivates.rb
```diff
@@ -0,0 +1,151 @@
...
+          else
+            uploaded_file(object)
+          end
+        end
+
+        def derivates_attribute
+          shrine_class.opts[:derivates_attribute]
+        end
+
+        def _set_derivates(uploaded_file)
```
janko-m: This method will always overwrite derivates the list. As discussed, we ideally want a method that can add/remove attachments to the current list.

lib/shrine/plugins/derivates.rb
```diff
@@ -0,0 +1,151 @@
...
+          instance_variable_defined?(:@old) || instance_variable_defined?(:@derivates_queue)
+        end
+
+        def finalize
+          super
+          update_derivates
+        end
+
+        def add_derivate(key, io, store = :store, nested_list = nil)
+          @derivates_queue ||= {}
```
janko-m: `@derivates_queue` should also handle removal of attachments. My vision is that I would like to make [revising derivates](https://shrinerb.com/rdoc/files/doc/regenerating_versions_md.html) easy, and that sometimes requires removing existing derivates.

lib/shrine/plugins/derivates.rb
```diff
@@ -0,0 +1,151 @@
...
+        def get
+          data = read
+          if data
+            if derivates_attribute == :original
+              data = JSON.parse(data)
+              data.delete "derivates" if data.has_key? "derivates"
+            end
+            uploaded_file(data)
+          end
+        end
```
janko-m: Note that `#read` might not always return a JSON string, when using JSON column with ActiveRecord/Sequel it will return a hash.

For now I would recommend this hack (until we think of a proper solution):

```rb
def get
  file = super
  file.instance_variable_set(:@data, file.data.reject { |key| key == "derivates" })
  file
end
```

hmistry:
Thanks for the quick feedback. Sounds like it's in the right direction and I can proceed with building it out. As per our earlier discussion, and now that I've properly reviewed your comments, there are many little aspects still to build and refine. I'll tackle them piece by piece and have discussions either here or in other mediums.

>If we manage to get derivates working in the same column, what would be the advantage of having a separate column?

At this point we don't seem to have a strong compelling reason to go one direction or the other. Yes, we can pick one and go with it, but this concept showed me how little code is needed to actually cater for both. Some people hate doing db migrations in case something goes wrong in production migration and others don't want to touch the original metadata in the event it gets corrupted while creating derivates. For now, I'll keep both in so we can try it out. If the code gets too complex or we find a good reason to go one way, it will be easy to do.

janko-m:
After an offline discussion, it seems that this will be getting rewritten, so I'll close the PR for now and let's open a new one when the new implementation is ready for review.

PR #300, opened 2018-10-01
Add a note regarding S3 performance on storage folder structure

hmistry:


janko-m:
Thanks for the PR!

I don't think it's necessary to have S3-specific guidelines in the `pretty_location` plugin documentation, especially because these are not related. I first thought they might be a little related, because Amazon S3 docs used to recommend inverting incrementing integers for better partitioning, but now this doesn't seem like a requirement anymore.

In the `Shrine::Storage::S3` documentation I would just note that there are per-prefix limits to req/s (maybe mentioning current limits), and recommend people to use prefixes if they need to scale further. 
Anyway, I think just loading the `pretty_location` plugin should do it.

hmistry:
All done. Thanks for the feedback. I was also hesitant about putting it in the `pretty_location` plugin because it's specific to a storage.

The `pretty_location` plugin creates a single prefix which is the `:model` name and as such will have the basic request rate limit. Also if the user creates a folder structure under `cache` and `store` in a bucket, they will limited to basic limit. The limits much higher than before but for some apps that scale, they will hit it.

PR #302, opened 2018-10-15
Add a code hint for Roda requirement to download endpoint docs

thomasklemm:
Hi Janko, thanks for this amazing gem and ecosystem! 

I've just run into a bit of trouble with less helpful looking error messages (though the end of the stacktrace points to the main reason) when adding the download endpoint plugin without having added the `roda` gem to the bundle first. Please decide if you want to add this hint, otherwise you can just close the PR.


janko-m:
Thanks for the patch!

Issue #303, opened 2018-10-23
mime_type cannot open: No such file or directory


kashifnaseer:
I'm using **determine_mime_type** and it works fine most of time but some times, it gives **error mime_type cannot open No such file or directory** while same file can be uploaded after multiple tries.  any idea what is causing this ?

```json
{
  "id": "f972ca5ea746dfc641f390757a3acb7b",
  "storage": "cache",
  "metadata": {
    "filename": "Oman Integrated Residential Project.jpg",
    "size": 111662,
    "mime_type": "cannot open: No such file or directory",
    "cloudinary": {
      "public_id": "cache/c",
      "version": 1540305195,
      "signature": "c",
      "resource_type": "raw",
      "created_at": "2018-10-23T14:33:15Z",
      "tags": [],
      "bytes": 111662,
      "type": "upload",
      "etag": "c",
      "placeholder": false,
      "url": "http://res.cloudinary.com/ccc/raw/upload/v1540305195/cache/c",
      "secure_url": "https://res.cloudinary.com/ccc/raw/upload/v1540305195/cache/c",
      "original_filename": "file"
    }
  }
}
```

shrine Initializer code for determine_mime_type:

```rb
require "cloudinary"
require "shrine/storage/cloudinary"
require "shrine/storage/s3"

Shrine.storages = {
  cache: Shrine::Storage::Cloudinary.new(store_data: true, prefix: "cache", resource_type: :raw), 
  store: Shrine::Storage::Cloudinary.new(store_data: true),
  s3_cache: Shrine::Storage::S3.new(prefix: :cache, access_key_id: aws_settings['k'], region: 'eu-west-1', secret_access_key: aws_settings['k'], bucket: aws_settings['b']),
  s3: Shrine::Storage::S3.new(access_key_id: aws_settings['k'], region: 'eu-west-1', secret_access_key: aws_settings['k'], bucket: aws_settings['b'])
}

Shrine.plugin :activerecord
Shrine.plugin :cached_attachment_data
Shrine.plugin :determine_mime_type
```

hmistry:
I'm not sure. Could you provide a backtrace so I can see where the fail is occurring?

One thought I have is I know in S3 there can be some delay before a file is actually ready after it's uploaded to it so depending on your architecture and timing, you could be trying to download the file before it's available. I'm not sure if Cloudinary has a similar thing.

janko-m:
[...]

PR #304, opened 2018-10-24
Fail when file command returned error in stdout

janko-m:
It appears that the file command can also fail with a successful exit status, in which case the error is returned in stdout. This commit handles that.

So far the only error we know about is

```
cannot open: No such file or directory
```

It's very odd this error happens, because Shrine passes file content to the standard input, it doesn't use a path on disk, so I don't know how this error could happen.

See https://github.com/shrinerb/shrine/issues/303

### DIFF
--- a/lib/shrine/plugins/determine_mime_type.rb
+++ b/lib/shrine/plugins/determine_mime_type.rb
@@ -179,9 +179,14 @@ class Shrine
 
             raise Error, "file command failed to spawn: #{stderr.read}" if status.nil?
             raise Error, "file command failed: #{stderr.read}" unless status.success?
+
             $stderr.print(stderr.read)
 
-            stdout.read.strip
+            output = stdout.read.strip
+
+            raise Error, "file command failed: #{output}" if output.include?("cannot open")
+
+            output
           end
         rescue Errno::ENOENT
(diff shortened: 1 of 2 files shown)


lib/shrine/plugins/determine_mime_type.rb
```diff
@@ -179,9 +179,14 @@ def extract_with_file(io)
 
             raise Error, "file command failed to spawn: #{stderr.read}" if status.nil?
             raise Error, "file command failed: #{stderr.read}" unless status.success?
+
             $stderr.print(stderr.read)
 
-            stdout.read.strip
+            output = stdout.read.strip
+
+            raise Error, "file command failed: #{output}" if output.start_with?("cannot open")
```
hmistry: Will this strict start with matching "cannot open" work across platforms/OS or would it better to match "cannot open" or "no file" anywhere in the output and downcase output?

janko-m: Yeah, we don't have to match at the start, maybe that will match more platforms. Between "cannot open" and "No such file", I don't know which one would be more cross-platform, since I don't know how messages will differ 🤷‍♂️

PR #311, opened 2018-11-07
Use a guard raise cause for `bucket` argument in S3

ardecvz:
Hi, Janko!

First of all, thank you for your a hard work on Shrine! It's a indispensable tool for complex upload workflows with a really sleek design.

Back to PR, I'd like to use a guard raise cause for bucket argument in S3 for an appropriate error message.

Before the change, it fails with a misleading error message mentioning
a missing `name` option (`missing required option :name`).

The raised error originated from S3 gem instead of Shrine.

It leads to a little confusion as a bucket is often configured with
an environment variable which is set by system administrators.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## HEAD
+
+* Use a guard raise cause for `bucket` argument in S3 for an appropriate error message (@ardecvz)
+
 ## 2.13.0 (2018-11-04)
 
 * Specify UTF-8 charset in `Content-Type` response header in `presign_endpoint` plugin (@janko-m)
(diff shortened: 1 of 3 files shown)


janko-m:
Looks good, thanks for the patch!

Yeah, this code made sense for when `Aws::S3::Resource#bucket` was returning `nil` when bucket name was `nil` in prior versions of aws-sdk-s3, but it seems that this changed and that aws-sdk-s3 now raises an exception (which is good).

> First of all, thank you for your a hard work on Shrine! It's a indispensable tool for complex upload workflows with a really sleek design.

That's really great to hear ❤️

PR #314, opened 2018-11-09
Updated S3 storage documentation

igor-alexandrov:
Require dependencies prior to initialising.


janko-m:
Thanks for the patch!

PR #323, opened 2018-11-19
Force-reload option for *_attacher methods

printercu:
Brings public api for reloading attachers.


janko-m:
Thanks! Could I ask what is the use case for this feature?

printercu:
1) in copy plugin
2) if someone wants to `image_attacher(store: custom_store) && do_smth`. Don't he ever want to call then `image_attacher(default_options)` to revert `store: :custom_store`?

PR #325, opened 2018-11-20
Don't define methods dynamically to not clear ruby's method cache

printercu:


janko-m:
@hmistry He said "define methods statically to **not** clear Ruby's method cache" 😉

hmistry:
Sigh! The double not's... too early I guess, the caffeine hasn't kicked in yet. 😃

PR #341, opened 2019-01-16
Extended determine MIME type with Marcel

skarlcf:
Marcel based on gem MimeMagic which has problems with some MS Office files: [".xls file detected as application/x-ole-storage"](https://github.com/minad/mimemagic/issues/50)

Marcel method 'for' method call with additional options 'name' resolve problem for tested files.

`Marcel::MimeType.for(pathname_or_io = nil, name: nil, extension: nil, declared_type: nil)`

### DIFF
--- a/lib/shrine/plugins/determine_mime_type.rb
+++ b/lib/shrine/plugins/determine_mime_type.rb
@@ -216,7 +216,7 @@ class Shrine
 
           return nil if io.eof? # marcel returns "application/octet-stream" for empty files
 
-          Marcel::MimeType.for(io)
+          Marcel::MimeType.for(io, name: extract_filename(io))
         end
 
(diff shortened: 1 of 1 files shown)


janko-m:
Looks good, thanks for the patch! Would you mind adding some tests for this change?

janko-m:
Merged in https://github.com/shrinerb/shrine/commit/b1e5d8362a8204e287d1609b580ccb90f8f009d6.

janko:
Just following up, this change was reverted in https://github.com/shrinerb/shrine/pull/356. You can still have that behaviour, but you have to opt in like this:

```rb
plugin :determine_mime_type, analyzer: :marcel, analyzer_options: { filename_fallback: true }
```

PR #356, opened 2019-03-01
Allow options to analyzers in determining_mime_type plugin

hmistry:
Reverts "Extended determine MIME type with Marcel" commit b1e5d836.

The change was introduced in v2.15.0. Since the filename can be anything, it was found in some small cases the mime-type determined based on filename can be incorrect. This also created a potential security issue for an application because filenames cannot be trusted. Therefore the change is reverted.

This commit allows passing in options to mime-type analyzers in `determine_mime_type` plugin. For now, Marcel is the only analyzer that makes use of any passed-in options but the implementation allows for other analyzers to have options if needed in future.

Marcel analyzer takes one option called `:filename_fallback` which can be set to true or false (default). When set to true, Marcel will also use the filename, if available, to determine the mime-type when it fails to find it using the file's magic number. This is designed to be opt-in for security reasons stated above.

Addresses issue #354.

### DIFF
--- a/doc/plugins/determine_mime_type.md
+++ b/doc/plugins/determine_mime_type.md
@@ -9,17 +9,19 @@ plugin :determine_mime_type
 
 By default the UNIX [file] utility is used to determine the MIME type, and the
 result is automatically written to the `mime_type` metadata field. You can
-choose a different built-in MIME type analyzer:
+choose a different built-in MIME type analyzer, for example:
 
 ```rb
 plugin :determine_mime_type, analyzer: :marcel
 ```
 
+## Analyzers
+
 The following analyzers are accepted:
 
 | Name            | Description                                                                                                                                                                                                                                                                       |
 | :------         | :-----------                                                                                                                                                                                                                                                                      |
-| `:file`         | (Default). Uses the [file] utility to determine the MIME type from file contents. It is installed by default on most operating systems, but the [Windows equivalent] needs to be installed separately.                                                                            |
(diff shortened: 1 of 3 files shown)


doc/plugins/determine_mime_type.md
```diff
@@ -28,8 +30,33 @@ The following analyzers are accepted:
...
+You'll need to ensure the file utility or gem is in installed on your system
+for Shrine to use the analyzer.
+
+## Analyzer Options
+
+Currently, Marcel is the only analyzer that accepts options to be passed in.
+Analyzer options can be set in one of two ways:
+
+```rb
+plugin :determine_mime_type, analyzer: :marcel, options: { filename_fallback: true }
```
janko: I think it would be more explicit if `:options` was renamed to `:analyzer_options`. What do you think?

lib/shrine/plugins/determine_mime_type.rb
```diff
@@ -13,15 +13,21 @@ def self.configure(uploader, opts = {})
         end
 
         uploader.opts[:mime_type_analyzer] = opts.fetch(:analyzer, uploader.opts.fetch(:mime_type_analyzer, :file))
+        uploader.opts[:analyzer_options] = opts.fetch(:options, uploader.opts.fetch(:analyzer_options, {}))
```
janko: Let's name the saved option `:mime_type_analyzer_options`, because `store_dimensions` plugin also has the concept of "analyzers", so if we decide to allow passing options there, then if we'd name it `:analyzer_options` as well, it would clash.

PR #357, opened 2019-03-02
Improve multi-chunk range requests

zarqman:
Range requests that began in a 2nd or later chunk were generating a requested range where the begin value was greater than the end value (eg: `22222..16383`). This caused `chunk.byteslice(...)` to return `nil` instead of a String, which in turn causes Puma (at least--others untested) to raise an exception.

This commit optimizes the begin/end condition to skip the chunk anytime the whole chunk is outside of range, which also avoids the possibility of yielding nil.

It also adds an optimization to end the entire iteration once the Down::ChunkedIO's position has gone past the requested range, which eliminates wasted effort downloading remote data that will be discarded anyway.


janko:
This looks good, thanks a lot for the patch! 👍

PR #361, opened 2019-03-05
Correct test description for determine_mime_type

aried3r:


test/plugin/determine_mime_type_test.rb
```diff
@@ -150,7 +150,7 @@
         assert_equal "application/octet-stream", @shrine.determine_mime_type(fakeio("😃", filename: "smile.jpeg"))
       end
 
-      it "returns application/octet-stream for unidentified MIME types" do
+      it "returns image/jpeg for unidentified MIME types when using filename fallback" do
```
janko: Technically it doesn't always return `image/jpeg`, it depends on the extension 😃. So maybe it could be "it uses the filename fallback for unidentified MIME types".

janko: Thanks for the patch, btw

aried3r: Yeah, I realize that, but I was describing this very test. But you're right, it should read something more generic. I fail to find a short description, because what it does is:
`it "returns the MIME type based on the filename when the filename fallback is deactivated and fails to detect the MIME type based on the files contents" do` 😅

hmistry: @aried3r Good catch. The test should say `it "returns the MIME type based on filename for unidentified MIME types when using filename fallback"`. It's not perfect but sufficient.

aried3r: > Good catch.

For RSpec there is a rubocop cop to detect this, for minitest I don't know.
https://rubocop-rspec.readthedocs.io/en/latest/cops_rspec/#rspecrepeateddescription

Anyway, I've changed the description!

PR #362, opened 2019-03-07
Add support to derivation endpoint for Rack < 2

Antsiscool:
Rack::Utils#escape_path has different behaviour for Rack < 2 and Rack >= 2.

Rack < 2 will escape '/' characters in the path while Rack >= 2 does not escape '/'.


hmistry:
@Antsiscool Thank you for the PR.

@janko: Would you mind reviewing this? It looks like a simple change to allow compatibility to Rack < 2 but I'm unsure if other rack-using plugins will work with Rack < 2 and how it might affect our support matrix. In our dev dependency, we require Rack >= 2.

PR #363, opened 2019-03-11
Refactor derivation endpoint UrlSigner

aglushkov:
UrlSigner was tightly coupled with internal implementation of Derivation::Url.
Now UrlSigner is a universal class, that accepts single `url` parameter.

This change opens possibility to refactor this code further
 - make UrlSigner a separate dependency
 - disable signing urls
 - providing custom UrlSigner

### DIFF
--- a/lib/shrine/plugins/derivation_endpoint.rb
+++ b/lib/shrine/plugins/derivation_endpoint.rb
@@ -308,13 +308,24 @@ class Shrine
       # serializes the source uploaded file into an URL-safe format
       source_component = source.urlsafe_dump(metadata: metadata)
 
+      # generate plain URL
+      url = plain_url(name, *args, source_component, params)
+
       # generate signed URL
-      signed_url(name, *args, source_component, params)
+      signed_url(url)
     end
 
-    def signed_url(*components)
+    def plain_url(*components, params)
+      # When using Rack < 2, Rack::Utils#escape_path will escape '/'.
+      # Escape each component and then join them together.
+      path = components.map{|component| Rack::Utils.escape_path(component.to_s)}.join('/')
+      query = Rack::Utils.build_query(params)
+      "#{path}?#{query}"
+    end
+
+    def signed_url(url)
       signer = UrlSigner.new(secret_key)
-      signer.signed_url(*components)
+      signer.signed_url(url)
     end
   end
 
@@ -706,19 +717,13 @@ class Shrine
       @secret_key = secret_key
     end
 
-    # Returns a URL with the `signature` query parameter generated from the
-    # given path components and query parameters.
-    def signed_url(*components, params)
-      # When using Rack < 2, Rack::Utils#escape_path will escape '/'.
-      # Escape each component and then join them together.
-      path = components.map{|component| Rack::Utils.escape_path(component.to_s)}.join('/')
-      query = Rack::Utils.build_query(params)
+    # Returns a URL with the `signature` query parameter
+    def signed_url(url)
+      signature = generate_signature(url)
+      query = Rack::Utils.build_query(signature: signature)
 
-      signature = generate_signature("#{path}?#{query}")
-
-      query = Rack::Utils.build_query(params.merge(signature: signature))
-
-      "#{path}?#{query}"
+      glue = url.include?('?') ? '&' : '?'
+      "#{url}#{glue}#{query}"
     end
 
     # Calculcates the signature from the URL and checks whether it matches the


aglushkov:
Personally I need this change to safely monkey patch UrlSigner to disable signing entirely.

I use derivation urls this way
```ruby
class ImageUploader < Shrine
  plugin :derivation_endpoint, secret_key: 'foobar'

  derivation(:'72x40x1') do |file|
    ImageProcessing::MiniMagick.source(file).resize_to_fill!(72, 40)
  end
end
```  

While I don't have user-provided params, I don't need signed urls.

janko:
Thanks for that initiative, disabling signatures was definitely on my TODO list for `derivation_endpoint`.

If we want to allow users to provide a custom signer, maybe we could simplify the signer even further, so that it only needs to generate a signature (and not have to concatenate query params), for example:

```rb
Shrine.plugin :derivation_endpoint, signature: -> (string) { ... return signature ... }
```

And for disabling signing just provide `false`:

```rb
Shrine.plugin :derivation_endpoint, signature: false
```

I don't know if you envision people wanting to customize signing further than just providing code which generates a signature (or disabling signatures). But you're right that it makes sense more for `UrlSigner#sign_url` and `UrlSigner#verify_url` to accept the same kind of parameters, so your PR improves that consistency.

These were just some initial thoughts I had.

> While I don't have user-provided params, I don't need signed urls.

Could you clarify this use case more? The signing is not only a protection for derivation args coming from user input, but it also prevents the user from attacks like generating a script that would hitting your server with expensive transformations that your app isn't even using (and avoiding the CDN cache).

aglushkov:
It's a great proposal to add `signature` option, I like it a lot

> &mdash; While I don't have user-provided params, I don't need signed urls.
> &mdash; Could you clarify this use case more? The signing is not only a protection for derivation args coming from user input, but it also prevents the user from attacks like generating a script that would hitting your server with expensive transformations that your app isn't even using (and avoiding the CDN cache).

I thought expensive transformations could be performed only by setting custom users params, like `/thumb/999999/999999`, I could miss something, but I thought signature just prohibits to request links that does not used in app

If I have just `/thumb` url without any param, client can request `/thumb/9999/9999`, but the result should be just 404 NotFound I think

aglushkov:
BTW, I use AWS Cloudfront as CDN, and it returns cached data regardless of provided GET params, so requests like `/thumb`, `/thumb?a=1`, `/thumb?a=2` should touch server once only

janko:
> I thought expensive transformations could be performed only by setting custom users params, like /thumb/999999/999999, I could miss something, but I thought signature just prohibits to request links that does not used in app

Yes, that's correct. I understand your approach now, define a derivation for each kind of possible parameter. Yeah, in that case it should be safe not to use a signature.

> If I have just /thumb url without any param, client can request /thumb/9999/9999, but the result should be just 404 NotFound I think

If you define a `:thumb` derivation but don't do anything with the arguments, a request to `/thumb/9999/9999` will still get in (as `derivation_enpdoint` doesn't currently check the arity of the derivation block), but you just won't use them in your block. However, it will bypass the CDN cache 🤔

> BTW, I use AWS Cloudfront as CDN, and it returns cached data regardless of provided GET params, so requests like /thumb, /thumb?a=1, /thumb?a=2 should touch server once only 

Cloudfront doesn't react to GET parameters? That's unfortunate, because the cache busting strategy of `derivation_endpoint` relies on adding query parameters (adding the `version` query parameter).

aglushkov:
> Cloudfront doesn't react to GET parameters?  That's unfortunate...

I think it's not big deal, version parameter can be whitelisted in Cloudfront https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/distribution-web-values-specify.html#DownloadDistValuesQueryString

janko:
I've struggled to make derivation URL arguments respect arity of derivation blocks, but I didn't manage to come to a clean enough solution. The problem is that derivation blocks are saved as procs (as opposed to lambdas), which have lax argument passing. It turns out it's difficult to convert a proc into a lambda, and even if we did so we might break backwards compatibility for anyone that was relying on lax arguments.

So, if we disabled signature, an attacker could still avoid the CDN cache by passing more URL arguments. Moreover, even if we solved that part, an attacker could add metadata to the serialized uploaded file in the URL, which would also bypass the cache.

I don't think it's possible to make the derivation endpoint safe without signatures, in which case I wouldn't feel comfortable supporting it. I would recommend creating your own derivation endpoint which does what you want, for example:

```rb
# config/routes.rb
Rails.application.routes.draw do
  # ...
  resources :photos do
    member { get "/thumbnail/:size", action: :thumbnail, as: :thumbnail }
  end
  # ...
end
```
```rb
# app/controllers/photos_controller.rb
class PhotosController < ApplicationController
  ALLOWED_SIZES = %w[72x40x1 ...]

  # ...
  def thumbnail
    head :unprocessable_entity unless ALLOWED_SIZES.include?(params[:size])

    photo = Photo.find(params[:id])
    set_rack_response photo.image.derivation_response(:thumbnail, *params[:size].split("x"))
  end

  private

  def set_rack_response((status, headers, body))
    self.status = status
    self.headers.merge!(headers)
    self.response_body = body
  end
  # ...
end
```
```rb
class ImageUploader < Shrine
  derivation :thumbnail do |file, width, height, scale|
    # ...
  end
end
```

------

That being said, this PR is about refactoring `UrlSigner`, and I generally agree with it. At first I wanted to push URL generation down into the signer so that other components can also potentially use it, but it breaks SRP (URL generation & generating signature). Also, this PR makes `#signed_url` and `#verify_url` methods symmetrical in terms of arguments.

I'll merge it shortly, sorry for the delay.

PR #367, opened 2019-03-15
Add  ignore errors option in remote url plugin

tanmen:
I'm sorry for the sudden PR.
I am currently using Shrine for a project. when using remote url, I thought that ignore errors was necessary for background and batch processing.
I expect this to be merge.

My English may be terrible because I am Japanese.

### DIFF
--- a/doc/plugins/remote_url.md
+++ b/doc/plugins/remote_url.md
@@ -83,6 +83,12 @@ plugin :remote_url, error_message: "download failed"
 plugin :remote_url, error_message: -> (url, error) { I18n.t("errors.download_failed") }
 ```
 
+## Ignore errors
+If you encounter an error in Download, use this to disable the error.
+```rb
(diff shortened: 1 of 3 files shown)


janko:
Thank you for the pull request. However, if you're assigning the file from a remote URL in a background job and want errors to be ignored, you should really just use the [Down](https://github.com/janko/down) gem directly, and catch the exceptions yourself. For example:

```rb
require "down"

begin
  file = Down.download(remote_url)
  photo.image = file
rescue Down::Error
  # ignore
end
```

The `remote_url` plugin is only useful in scenarios where you want to give the user the ability to select a file from a remote URL instead of uploading their own file. In this case you want to have some validation errors if downloading from the URL fails.

tanmen:
Thank you for a reply.
I understood that it was code that did not fit this plug-in.
I decided to handle errors on my own.

PR #371, opened 2019-04-02
Versions: allow Attacher#url to accept version name indifferently

FunkyloverOne:
So both String & Symbol version names are working.

Perhaps this implementation could be "stricter" by simply symbolizing input version without "pre-symbolized" try.
I did it the way it is because I believe this approach is potentially less buggy, all it does is additionally tries symbolized version afterward in case the given one did not work out.

### DIFF
--- a/lib/shrine/plugins/versions.rb
+++ b/lib/shrine/plugins/versions.rb
@@ -105,7 +105,13 @@ class Shrine
 
           if attachment.is_a?(Hash)
             if version
-              indifferent_version_url(version, **options)
+              if attachment.key?(version)
+                attachment[version].url(**options)
+              elsif fallback = shrine_class.version_fallbacks[version]
+                url(fallback, **options)
+              else
+                default_url(**options, version: version)
+              end
             else
               raise Error, "must call Shrine::Attacher#url with the name of the version"
             end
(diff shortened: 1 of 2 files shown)


FunkyloverOne:
Whoopsie, something is still not quite right with JRuby

lib/shrine/plugins/versions.rb
```diff
@@ -105,13 +105,7 @@ def url(version = nil, **options)
 
           if attachment.is_a?(Hash)
             if version
-              if attachment.key?(version)
```
janko: This solution seems a bit complex, wouldn't just adding

```rb
version = version.to_sym
```

above this line work?

FunkyloverOne: @janko yup, it will work indeed. Though my solution won't break anything in case any existing version name is for any reason not a Symbol.
If you feel like that could never happen - then your simpler solution fits better :)

FunkyloverOne: @janko You see, we're only symbolizing version names in `Shrine.uploaded_file`, which converts a hash of data into a hash of versions, mostly during the read from DB I believe.

But what if that hash of versions was somehow modified afterward?
Can we guarantee all the keys are going to be symbols? :thinking: 

I'm just a bit paranoic you see...

FunkyloverOne: Anyway, I didn't want to introduce potential bugs with this my little enhancement.

janko: Thanks for being thorough, and thanks for the PR in general! I think it would be very strange if a Shrine user intercepted reading from the DB and added some versions with string keys, so I believe the approach I suggested shouldn't cause any problems.

FunkyloverOne:
@janko agreed! Here we go, now it's simple and clean :)

janko:
Thank you, merging 👍

PR #378, opened 2019-04-27
add bundle exec to sequel in demo README

hide-furusawa:


hmistry:
Did you have an issue running the command without `bundle exec`?

hide-furusawa:
I got error. and I thought it would be better to match it with the rackup command to README.

janko:
> I got error

Which error?

janko:
I'm going to merge this, as I don't see any harm in adding `bundle exec`, and I trust you that you had problems without. I was just curious what problem it was, so that I increase my understanding of Ruby. Thanks for the patch!

hide-furusawa:
Thank you for the reply and merge.
I will use shrine .

PR #380, opened 2019-04-29
Prevent sharing of process blocks between parent and subclasses

ksol:
Cf #379 for the full issue. Long story short, blocks registered via `process` were not behaving as one would expect when defined accross a hierarchy of Uploaders.


janko:
Looks good, thanks for the patch! I’ll merge tomorrow.

PR #385, opened 2019-05-19
Add instructions for vanilla setup

mariokostelac:
While setting up shrine for the first time, it took longer than expected to figure out how to set it up to "just upload a file". I believe this makes it easier.

_Thanks for creating this gem!_


janko:
Thanks for the patch! I agree it's worth mentioning, I often don't bother creating an uploader when creating a simple script that does something with Shrine.

PR #390, opened 2019-06-22
Fix typo in S3 documentation

sjaq:
Accidentally copy pasted this without noticing the mistake and ran into configuration issues.


janko:
Thanks!

PR #392, opened 2019-06-28
Allow setting the identifier in pretty location plugin

00dav00:
This PR allows `pretty_location` plugin to use different identifiers when building the location.
[Here](https://groups.google.com/forum/#!topic/ruby-shrine/7Wm1B529I30) is the discussion where this was proposed.

### DIFF
--- a/doc/plugins/pretty_location.md
+++ b/doc/plugins/pretty_location.md
@@ -28,7 +28,7 @@ plugin :pretty_location, namespace: "/"
 # "blog/user/.../493g82jf23.jpg"
 ```
 
-By default if a record responds to `id`, this will is used in the location. 
+By default, if there is a record present, the record `id` will is used in the location.
 If you want to use a different identifier for the record, you can pass in
 the `:identifier` option with the desired method/attribute name as the value:
 
@@ -39,5 +39,15 @@ plugin :pretty_location, identifier: "uuid"
 plugin :pretty_location, identifier: :email
 # "user/foo@bar.com/profile_picture/493g82jf23.jpg"
 ```
+
+For a more custom identifier logic, you can overwrite the method `generate_location`
+and call `pretty_location` with the identifier you have calculated.
+
+```rb
+def generate_location(io, context)
+  identifier = context[:record].email if context[:record].is_a?(User)
+  pretty_location(io, context, identifier: identifier)
+end
+```rb
 
 [pretty_location]: /lib/shrine/plugins/pretty_location.rb
--- a/doc/plugins/pretty_location.md
+++ b/doc/plugins/pretty_location.md
@@ -28,4 +28,16 @@ plugin :pretty_location, namespace: "/"
 # "blog/user/.../493g82jf23.jpg"
 ```
 
+By default if a record responds to `id`, this will is used in the location. 
+If you want to use a different identifier for the record, you can pass in
+the `:identifier` option with the desired method/attribute name as the value:
+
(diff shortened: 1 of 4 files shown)


janko:
Thanks for the PR!

I would also like to handle the case where the same uploader is used for models that do and don't have the other identifier column. For example, if the user wants to use an `ImageUploader` for a `User.avatar` and a `Photo.image`, and for `User.avatar` they want to use the user's `email` but for `Photo.image` they want to use the default `id`.

In our discussion I proposed moving the pretty location implementation into an instance method that you can call from `#generate_location`. I feel like that approach could solve the above scenario nicely.

00dav00:
I see, let me take a look at this @janko

00dav00:
@janko Is this more like what you have in mind?

janko:
Yes, this is more in line with what I wanted, thank you. I will make some minor changes before merging, such as the identifier fallbacks which I meant for users to handle themselves, but the main parts look good 👍

I should have time some time tomorrow to merge.

00dav00:
Awesome! I created the default identifier logic so that if you set the identifier to be the `uuid` to be used in an uploader and you use it in a model that does not respond to this method, the ``

00dav00:
Hey @janko ! Any news on this one?
If you give me some pointers on the changes have in mind I would be have to do them if you don't have the time now. Please let me know :)

janko:
Hey @00dav00, sorry for the delay, and thank you for your patience. Let's get this merged!

First, I would like the `#pretty_location` method to accept the `identifier` as a keyword argument, so that the user can call it like this:

```rb
def generate_location(io, context)
  pretty_location(io, context, identifier: context[:record].email)
end
```

This will open space for potentially more options in the future.

Also, I like the plugin itself not to do any identifier fallbacks, as I feel that's out of scope for Shrine. If the user needs them, they can implement them themselves by overriding `#generate_location`.

Finally, it would be nice to update the documentation showing that `#pretty_location` method can be called when overriding `#generate_location`. E.g. the documentation example could show an example of identifier fallbacks like this one:

```rb
def generate_location(io, context)
  identifier = context[:record].email if context[:record].is_a?(User)
  pretty_location(io, context, identifier: identifier)
end
```

After that I'll merge it right way.

00dav00:
Awesome, I will do those changes in a bit, just one question

Let's say we have two models: `User(:id, :email, :profile_pic)` & `Post(:id, :uuid, :image, :text)`

And we set the initializer to work like this:
```ruby
plugin :pretty_location, identifier: :uuid
```
If we remove the default identifier logic I think we have a couple options:
 - Raise a method not found error when the `User` model tries to call uuid in the uploader
 - Pass `nil` as the identifier making the route looks like `user/profile_pic/78127391823.jpg`

Which one do you think would be the best option here @janko?

janko:
I think raising an exception would be more appropriate.

00dav00:
Hey @janko , just pushed the changes.
I kept the method `record_identifier` because it helps making the `identifier` set optional by trying to call `id` when the option is not present in the initializer.
I'm checking for the record presence inside `record_identifier`t to [keep this test](https://github.com/shrinerb/shrine/blob/2fb4923bbfec774c9d3077c2e85b827ae8660d73/test/plugin/pretty_location_test.rb#L41) green.

WDYT?

janko:
@00dav00 Thanks, looks good, I will be merging tomorrow.

PR #393, opened 2019-07-01
Update docs: env deprecated in rails 5. use request.env

jalevin:


janko:
Thanks for the patch!

PR #394, opened 2019-07-02
Small grammar correction in README

preciz:


janko:
Thanks for the patch!

PR #398, opened 2019-07-23
Fix activerecord plugin's after_commit callbacks to work in Rails 3.

texpert:
Because the transaction_include_action? method in Rails 3  isn't supporting yet arrays of actions - see the code: https://github.com/rails/rails/blob/3-2-stable/activerecord/lib/active_record/transactions.rb#L370

### DIFF
--- a/lib/shrine/plugins/activerecord.rb
+++ b/lib/shrine/plugins/activerecord.rb
@@ -35,9 +35,11 @@ class Shrine
               attacher.save if attacher.changed?
             end
 
-            model.after_commit if: :persisted? do
-              attacher = send("#{name}_attacher")
-              attacher.finalize if attacher.changed?
+            [:create, :update].each do |action|
+              model.after_commit on: action do
+                attacher = send("#{name}_attacher")
+                attacher.finalize if attacher.changed?
+              end
             end
 
             model.after_commit on: :destroy do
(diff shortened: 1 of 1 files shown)


texpert:
@janko, please, take a look! I would be grateful if this could be merged into a new release!

lib/shrine/plugins/activerecord.rb
```diff
@@ -35,12 +35,12 @@ def included(model)
               attacher.save if attacher.changed?
             end
 
-            model.after_commit on: [:create, :update] do
+            model.after_commit if: :persisted? do
```
janko: I'm not so confident this will work with any future version of Rails. Does `#persisted?` surely return `false` when the record is destroyed?

I think it would be safer to register one `after_commit` callback per action:

```rb
[:create, :update].each do |action|
  model.after_commit on: action do
    attacher = send("#{name}_attacher")
    attacher.finalize if attacher.changed?
  end
end
```

texpert: Yep, overnight and along the day the doubt was itching my mind too :)

Fixed in https://github.com/shrinerb/shrine/pull/398/commits/dcef9108c75de9c49bbf1bd2ea18e94c21f63c62

janko:
Version 2.19.2 released with this fix.

texpert:
Thanks a lot, @janko !!

PR #406, opened 2019-09-01
Improve the derivatives guide

hmistry:
Overall I like how this guide has turned out. It nicely addresses the different aspects of managing derivatives. In another guide, we can talk about background processing and parallelization strategies for generating derivatives so it doesn't convolute this one.

Have a few comments on this one which I'd like your thoughts on.

Introduce vs adding derivatives title - I see your intent and it's a good idea. I'd like to see if we can be a little more clearer because introduce can mean like an introduction and adding can also mean adding to original file. One thought was calling it "introduce derivatives to original attachment" and "adding more derivatives".

Generating URL's - we talk of you cannot generate url's until all existing derivatives are created but then we don't come back and show or talk about how to generate urls.

Removing derivative - if the attachment has changed, don't we need to still delete the `x_large` file because the error will be raised on `atomic_persist` and so it will not delete it?

### DIFF
--- a/doc/changing_derivatives.md
+++ b/doc/changing_derivatives.md
@@ -1,11 +1,14 @@
-# Changing Derivatives
+# Managing Derivatives
 
-This guide provides tips for modifying [derivatives] logic for an app that is
-already live and handling file attachments, with zero downtime.
+This guide shows how to add, create, update, and remove [derivatives] for an 
+app in production already handling file attachments, with zero downtime.
 
-The examples will be showing image thumbnails (`Photo` model with `image`
-attachment), but the advice applies to any kind of derivatives. We'll also be
-using the [Sequel] ORM, but it should easily translate to Active Record.
+*Note: The examples uses the [Sequel] ORM, but it should easily translate to 
+Active Record.*
+
+Let's assume we have a `Photo` model with an `image` file attachment. The 
+examples will be showing image thumbnails, but the advice applies to any kind 
+of derivatives. 
 
 ```rb
 Shrine.plugin :sequel
@@ -34,7 +37,12 @@ end
 *Scenario: Your app is currently working only with original files, and you want
 to introduce derivatives.*
 
-Start by loading the `derivatives` plugin and defining your processing logic:
+#### 1. Add `derivatives` plugin and `image_processing` gem
+
+Start by loading the `derivatives` plugin in the uploader and the 
+`image_processing` gem, and then define the processing logic. In this example, 
+we are creating a `thumbnails` derivatives processor which will be responsible 
+to create `small`, `medium`, and `large` thumbnails of the original `image`.
 
 ```rb
 # Gemfile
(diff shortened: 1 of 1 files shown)


doc/changing_derivatives.md
```diff
@@ -34,7 +37,12 @@ end
 *Scenario: Your app is currently working only with original files, and you want
 to introduce derivatives.*
 
-Start by loading the `derivatives` plugin and defining your processing logic:
+#### 1. Add `derivatives` plugin and `image_processing` gem
```
janko: I would rather that this guide focused on what to do _after_ a change in processing was made. This is how other sections in this guide are structured, and it makes sense to me to have it more conversational, because this guide covers several scenarios.

doc/changing_derivatives.md
```diff
@@ -34,7 +37,12 @@ end
...
+
+Start by loading the `derivatives` plugin in the uploader and the 
+`image_processing` gem, and then define the processing logic. In this example,
```
janko: The `image_processing` gem would not be needed if the user is not processing images. So it would be ideal if wording was more generic here.

doc/changing_derivatives.md
```diff
@@ -59,12 +67,17 @@ class ImageUploader < Shrine
...
+
+We can deploy this change to production.
 
-Next, we want to generate derivatives for all existing attachments in
-production. We can do this by running the following script:
+#### 2. Generate derivatives for existing attachments
+
+To generate derivatives for all existing attachments in production, run
+the following script. It fetches the photos in batches, downloads the image 
+in `:store` storage, creates derivatives, and persist the changes.
```
janko: I would prefer using the more high-level "permanent storage"  term here, so that it explains the conditional.

doc/changing_derivatives.md
```diff
@@ -90,8 +103,10 @@ Now all attachments should have correctly generated derivatives
 *Scenario: The processing logic has changed for all derivatives, and now you
 want to reprocess them for existing attachments.*
 
-Once the processing logic change has been deployed, we can run the following
-script to reprocess all derivatives:
+Deploy the processing logic change to production and then run the following
+script to reprocess all derivatives. It fetches the photos in batches, 
+downloads the image in `:store` storage, reprocesses new derivatives, persist
```
janko: Here as well.

doc/changing_derivatives.md
```diff
@@ -240,10 +264,10 @@ Photo.paged_each do |photo|
   x_large = attacher.remove_derivative(:x_large)
 
   begin
-    attacher.atomic_persist # persist changes if attachment has not changed in the meantime
+    attacher.atomic_persist       # persist changes if attachment has not changed in the meantime
     x_large.delete
-  rescue Shrine::AttachmentChanged, # attachment has changed
-         Sequel::NoExistingObject   # record has been deleted
+  rescue Shrine::AttachmentChanged,   # attachment has changed
+         Sequel::NoExistingObject     # record has been deleted
```
janko: Alignment seems a bit off here.

janko:
> It nicely addresses the different aspects of managing derivatives. In another guide, we can talk about background processing and parallelization strategies for generating derivatives so it doesn't convolute this one.

Sounds good to me 👍 

> Introduce vs adding derivatives title - I see your intent and it's a good idea. I'd like to see if we can be a little more clearer because introduce can mean like an introduction and adding can also mean adding to original file. One thought was calling it "introduce derivatives to original attachment" and "adding more derivatives".

Yeah, I see your point. Maybe this one can be called "adding derivatives", and the other one where we add new derivatives could be named "adding more derivatives"?

hmistry:
I've incorporated all your feedback. Thank you for your inputs!

Only thing left is the generate url part. I'm a bit unclear as to what attachment url's you are referring to, the derivative or image? And what should the user do for url's before all the processing is ready?

janko:
@hmistry Thank you for addressing!

> Only thing left is the generate url part. I'm a bit unclear as to what attachment url's you are referring to, the derivative or image?

The derivative URLs. The original file URLs are staying as they are. I'd like to avoid any code examples, as I don't want to suggest how and where they're using these URLs. But maybe we can mention something like "now you can update your application code to use new derivative URLs" or something. This is what you had in mind, right?

> And what should the user do for url's before all the processing is ready?

The guide implies that user's app was functioning before these derivatives changes. So the existing controller/model/view should be just kept as is, until all derivatives have been generated. Does that make sense?

PR #434, opened 2019-12-02
Add highlights and improve document landing page

hmistry:
I thought it would be nice to add a highlights section and a couple more code samples to the document landing page.

What do you think?

### DIFF
--- a/website/demo.md
+++ b/website/demo.md
@@ -1,6 +1,27 @@
-You can attach files...
+### Highlights
+
+* Modular and memory-friendly design
+* Storage options from disk to cloud and from single to multiple storages
+* Persistence integrations for a variety of ORM's with validation
+* Works with all the web frameworks - Roda, Rails, Hanami, Sinatra, etc
+* Configurable architecture from simple file upload to direct upload, eager or on-the-fly processing, resumable uploads, background processing, etc
+
+### Code Sample
+
+#### Setup the storages ...
 
 ```rb
+Shrine.storages = {
+  cache: Shrine::Storage::S3.new(prefix: "cache", **s3_options), # temporary
+  store: Shrine::Storage::S3.new(**s3_options),                  # permanent
+}
+```
+
+#### Easily attach files ...
+
+```rb
+# using :sequel plugin
+
 class Photo < Sequel::Model
   include Shrine::Attachment(:image)
 end
@@ -16,9 +37,11 @@ photo.image.storage  #=> #<Shrine::Storage::S3>
 photo.image.metadata #=> { "size" => 749238, "mime_type" => "image/jpeg", "filename" => "nature.jpg" }
 ```
 
-...and process them eagerly...
+#### Process them eagerly...
 
 ```rb
+# using :derivatives plugin
+
 Shrine::Attacher.derivatives :thumbnails do |original|
   magick = ImageProcessing::MiniMagick.source(original)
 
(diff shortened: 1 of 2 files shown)


janko:
Thanks for the PR, but I don't think this is an improvement to the landing page.

I tried adding similar highlights to the landing page in form of a bulleted list in the past, but it just turned out to be too much text. It could work if we put them in form of icons that are distributed in a grid ([example](https://ttytoolkit.org/)).

I also don't think that showing storage configuration helps, as I consider it to be the less important part, especially because it shows existence of temporary storage which I consider to be an implementation detail.

Similar for plugins, they are more of an implementation detail to manage complexity, I don't think it's useful to mention them in the illustration as it's not a known concept. My goal was to just show main features with as little setup as possible.

hmistry:
@janko Thank you for reviewing and feedback. I may not agree with your take on whether it's an improvement though I understand your point of view. 🙂 

The current landing to me does not speak or highlight what Shrine can do as people can reach the site directly and are tossed into 3 code samples without an intro/context. As such, I think adding a few bullets highlighting Shrine features improves that significantly, especially needing to differentiate it from other file attachment gems. It would be nice to have icons and formatting like your example but in the mean time, a few bullets won't hurt.

As for storage and plugins, you make a point. I wanted to showcase Shrine's strength and configurability to give a glimpse of what might be possible because these aspects of Shrine are equally valuable when compared to other file libraries. This is debatable and the above form may not be a good way.

I'll punt on this for now and will try again when I come up with some ideas.

janko:
Thanks for the feedback and for pushing for this.

> The current landing to me does not speak or highlight what Shrine can do as people can reach the site directly and are tossed into 3 code samples without an intro/context.

Yes, I totally agree with this, I'm also not happy with just having these code examples. I would like to have some additional illustrations:

1. I think it would be useful to present how the uploaded file data is stored, i.e. that it's JSON data stored in a `<attachment>_data` column. I was thinking of presenting this visually alongside the 1st code example in a separate column (Docusaurus has a [GridBlock component](https://docusaurus.io/docs/en/api-pages#complibrarygridblock) that might help with this), but it can also be in another way.

2. It would be nice to visually present eager vs on-the-fly processing with an image. I was thinking something like this for eager:

   ```
     + ---------------- +
     |                  |
     |                  |         + ------------ +
     |                  |         |              |
     |    some image    |         |              |              + ---------- +
     |                  |         |  some image  |              |            |
     |                  |         |              |              | some image |
     |                  |         |              |              |            |
     + ---------------- +         + ------------ +              + ---------- +
   https://s3.amazonaws.com/    https://s3.amazonaws.com/     https://s3.amazonaws.com/
   path/to/large.jpg            path/to/medium.jpg            path/to/small.jpg
   ```

   and the same for on-the-fly with derivation URLs instead. Just something that would visualize the code, though there might be better options of course.

> As such, I think adding a few bullets highlighting Shrine features improves that significantly, especially needing to differentiate it from other file attachment gems.

I agree I would like to somehow differentiate Shrine from alternatives, but for me personally bullet points didn't do anything, and just pushed the code example further down. I don't know what would be a good alternative, but I agree we should add some comparison somehow.

hmistry:
I like your ideas, especially visuals to show architecture. I think we can come up with something that satisfies both our concerns - will take some iterations. Stay tuned. 🙂

PR #445, opened 2020-01-23
Update carrierwave migration doc

pn-y:
Hi

currently, I'm migrating from carrierwave to shrine, and met an issue with not respected prefix during dual write. So I've decided to make some workaround to help.


janko:
Looks good, thanks for the patch!

PR #446, opened 2020-01-28
fix IO validation on upload_endpoint

pldavid2:
Fix IO validation on upload_endpoint as discussed here
https://discourse.shrinerb.com/t/upload-endpoint-with-authorization/121

Helps for example when someone wants to call Shrine.upload_response directly from a rails controller to add authentication.


janko:
The functionality looks good, thanks for the pull request! I will add tests and fix formatting, then it should be ready to merge.

PR #452, opened 2020-03-23
Update CoC to v2.0

hmistry:
Adds enforcement guidelines amongst other edits.

We recently updated CoC for Bundler due to feedback from few individuals. I thought it will be good to do the same for Shrine.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -60,7 +60,7 @@ representative at an online or offline event.
 
 Instances of abusive, harassing, or otherwise unacceptable behavior may be
 reported to the community leaders responsible for enforcement at
-[INSERT CONTACT METHOD].
+janko.marohnic@gmail.com.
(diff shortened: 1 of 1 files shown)


CODE_OF_CONDUCT.md
```diff
@@ -1,22 +1,128 @@
...
+an individual is officially representing the community in public spaces.
+Examples of representing our community include using an official e-mail address,
+posting via an official social media account, or acting as an appointed
+representative at an online or offline event.
+
+## Enforcement
+
+Instances of abusive, harassing, or otherwise unacceptable behavior may be
+reported to the community leaders responsible for enforcement at
+[INSERT CONTACT METHOD].
```
janko: Should we replace this with an email address? I'm ok if we just put my email address here.

hmistry: Yes, good catch!

PR #453, opened 2020-03-23
Update CONTRIBUTING.md

BKSpurgeon:
Add: dependency for libvips when running the test suite.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -41,6 +41,7 @@ The test suite requires that you have the following installed:
 
 * [libmagic]
 * [SQLite]
+* [libvips and libvips-dev] - please download the appropriate package suiting your operating system.
 
 If you're using Homebrew, you can just run `brew bundle`. The test suite is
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -41,6 +41,7 @@ The test suite requires that you have the following installed:
 
 * [libmagic]
 * [SQLite]
+* [libvips and libvips-dev] - please download the appropriate package suiting your operating system.
```
hmistry: Do you have to download both or just one depending on the OS? If latter, then "or" is more correct than "and".

I think it's better to simplify it to only "libvips", the most common one, and leave it up to the dev for their specific OS. Any thoughts?

BKSpurgeon: Sure, that makes sense. The tests will work just fine without  libvips-dev. I'll remove and update the link straight to the "how to install my build" page.

BKSpurgeon:
I made a new PR. So will close this one.

PR #454, opened 2020-03-24
Fix typo

BKSpurgeon:
Please confirm that the above is indeed a typo: I could not find a module called Storages (plural): Shrine::Storages so I made a big assumption that it should be Storage (singular)


hmistry:
Good catch! Thanks

PR #455, opened 2020-03-24
Add: instructions to download libvips in order to run testing suite

BKSpurgeon:


BKSpurgeon:
#453 - made the edits referred to here.

Hope it helps.

BKSpurgeon:
> Thanks for the PR +1 I'm thinking whether it would be better to instead skip tests that use libvips when libvips is not installed. I think that might lower the barrier for new contributors. What do you think?

My two cents:  

### The libvips issue

Personally speaking I don't think that `libvips` is or will be a limiting factor in contributions: people can easily (?) download libvips (provided that the instructions are clear) -- unless I'm missing something? 

Some ideas:

- Perhaps i can further clarify the download and installation instructions? 

- also to consider: how much effort will it take to restructure the tests vs. how much easier will that make it for new contributors?

Pls LMK what you wanna do here.

### The barrier to entry issue:

Fundamentally, I think that, as a new comer (and speaking from my own personal experience) it is not easy trying to grok and understand a new code base, and to understand how everything fits together in Shrine. if i can understand it, perhaps i might try to write a tutorial to hopefully help people quickly grok it.

Pls let me know your thoughts on the above. 

rgds
Ben

hmistry:
@janko I agree with @BKSpurgeon that adding `libvips` is not a big effort as it installs easily. I would worry more about test escapes if we start skipping tests because a dependency is not installed. 

@BKSpurgeon FYI, you could have pushed the changes to the existing PR #453.
If you have any ideas on how to improve "how everything fits together in Shrine", let us know in our discourse. We've tried our best to have guides and documentation and examples but there's a lot you can do with Shrine with all the plugins so it's a matter of discovering the right one.

No action required here. I'll merge this PR in a bit.

BKSpurgeon:
@hmistry agreed, the documents are amazing, bar none! I also will certainly advise via the discourse on my thoughts if and when they are more completely formed.

PR #462, opened 2020-04-27
Add validation helper that check mime type with regexp

y-yagi:
This is useful to check only top level type.
Also, similar libraries(Paperclip, CarrierWave) supports same format
validation. So this useful too when migration from other libraries.


janko:
Thank you for your pull request. However, Shrine already had this functionality built in, but I had removed it because it encourages accepting too wide range of image formats, which increases risks of vulnerabilities such as ImageTragick.

Note that you can still have this functionality by doing a custom validation:

```rb
Attacher.validate do
  if file.mime_type !~ /\Aimage\/.+\z/
     errors << "must be an image"
  end
end
```

PR #467, opened 2020-05-11
Add GCS expires not to url_options

M-Dahab:
For google cloud storage, the option is `expires` instead of S3's `expires_in`.

### DIFF
--- a/doc/plugins/url_options.md
+++ b/doc/plugins/url_options.md
@@ -10,6 +10,8 @@ storages.
 plugin :url_options, store: { expires_in: 24*60*60 }
 ```
 
+If you're using [GCS](https://googleapis.dev/ruby/google-cloud-storage/latest/Google/Cloud/Storage/File.html#signed_url-instance_method) you need `expires:` instead.
+
 You can also generate the default URL options dynamically by using a block,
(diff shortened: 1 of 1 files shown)


doc/plugins/url_options.md
```diff
@@ -10,6 +10,8 @@ storages.
 plugin :url_options, store: { expires_in: 24*60*60 }
 ```
 
+If you're using [GCS](https://googleapis.dev/ruby/google-cloud-storage/latest/Google/Cloud/Storage/File.html#signed_url-instance_method) you need `expires:` instead.
```
hmistry, line 13: This is a good attempt though I can see this cascading into adding Azure, and whatever other storage services people use which becomes a maintenance nightmare. Also how many other spots in the documentation will this happen.

While it is nice to have the referring links like this, I think the better solution will be to state that `expires_in` is an example specific to AWS S3 and that the reader should refer to their cloud service for appropriate keys.

One example is:

```
plugin :url_options, store: { expires_in: 24*60*60 }  # `expires_in` is a URL option for AWS S3
```

Thoughts?

M-Dahab:
Closed in favor of #469

PR #476, opened 2020-05-18
Enable options to trigger_mirror_upload to make e.g. resource and derivative name available

corneverbruggen:
This change allows options to be passed to trigger_mirror_upload, and then to upload. This way you can make e.g. the record en derivative name available to mirror upload. When combining this with the upload options plugin, you can make e.g. the record available there, for using it set S3 acl's.


janko:
Thank you for the patch, this looks good 👍 I will update tests and merge afterwards.

PR #483, opened 2020-06-16
Fix: comma included in hash

BKSpurgeon:
saves debugging for copy-pasters.


janko:
Ruby supports trailing commas in hashes and arrays, and I encourage people to use them (Shrine uses them as well), because they make editing easier and lead to better git histories.

Are there any problems with using them in examples?

BKSpurgeon:
> Ruby supports trailing commas in hashes and arrays, and I encourage people to use them (Shrine uses them as well), because they make editing easier and lead to better git histories.

I didn't know this. i made a mistake: I assumed that it could not be done.

PR #475, opened 2020-05-16
Add: instructions on how to run ruby templates

BKSpurgeon:
### What is this PR?

- New contributors might not know how to easily run the debugging templates (it took me a not insignificant time to learn how to do it). This PR adds instructions on how this can be done. Hopefully it will be of some use: if not, feel free discard it according to your discretion.

### What's the benefit of this PR?

* Hopefully this will lower the barriers to entry to new comers to help debug tests, and that will hopefully ease the maintenance burden for maintainers.

PR #486, opened 2020-06-18
Add: self contained example templates via a repository

BKSpurgeon:
### What's the benefit of this PR?

* This PR hopes to make it even easier for developers to get a simple self-contained example of potential problems which may exist. Hopefully this will ease the burden of the maintainers of this Gem in debugging, and help encourage folks to deliver more self-contained examples. Relevant discussion  #475.

### What are the changes made and why?

* I have added a [simple repository where people can download a self-contained example](https://github.com/BKSpurgeon/debugging-shrine-example) and modify it in less than 5 minutes. Just a simple `git clone` and running the ruby script is all that is needed now - before folks would have to cut and paste templates which might be cumbersome for them. 

* I have created a separate document for instructions re: self-contained example - so people can cut straight to the relevant part.

### If not a good idea, then feel free to close

* Decisions have to be made. Some ideas may not be good, including this one.

Hope it helps.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,12 +2,10 @@ Issue Guidelines
 ================
 
 1. Issues should only be created for things that are definitely bugs.  If you
-   are not sure that the behavior is a bug, ask about it on the [forum].
+   are not sure that the behavior is a bug, ask about it on the [forum]. Otherwise Github gets overwhelmed with issues and it is very difficult for the maintainers to manage.
 
 2. If you are sure it is a bug, then post a complete description of the issue,
-   the simplest possible self-contained example showing the problem (see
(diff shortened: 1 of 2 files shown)


hmistry:
Thanks for the PR. I like this idea and am for it. @janko what are your thoughts on this? 

Also thank you for making the separate git repo, I think it is helpful as git cloning is easier than copy-paste. I'm very curious to see if people will start using it because we try to encourage using the templates.

janko:
I like this as well, I think these kind of changes are exactly what we need to make Shrine easier to contribute to, so thank you 👍

hmistry:
Merged. Thank you for thinking about ways to make it easier for new and existing users. Keep sending PR's or suggest in our forum, we welcome your inputs. 🙂

PR #501, opened 2020-10-09
Update .travis.yml

gogainda:


janko:
Thanks, but I think you only meant to add TruffleRuby, without removing other Rubies.

gogainda:
Correct , I will fix the code change

Issue #503, opened 2020-10-27
Exception while trying to marshal ActiveRecord/Sequel model, with attachment & derivatives: no _dump_data is defined for class Thread::Mutex


Morozzzko:
### Brief Description

My app has a case when we need to cache an ActiveRecord object, which has a Shrine attachment.

The attachment's uploader works with two plugins: `activerecord` and `derivatives`.

This worked fine under 2.19.4, but broke after our upgrade to 3.3.0

### Expected behavior

I expect Rails cache to marshal the provided object and put in the cache. I also expect the object to be functioning properly afterwards

### Actual behavior

Whenever I try to marshal a newly-created object with an attachment, I get an exception: `no _dump_data is defined for class Thread::Mutex (TypeError)`

However, if I try to re-fetch the same object and marshal it, it works fine. `#reload` doesn't work, though.

### Simplest self-contained example code to demonstrate issue
Provide a self-contained example code (see the [Sequel/ActiveRecord templates](https://github.com/shrinerb/shrine/blob/master/CONTRIBUTING.md)) to show the problem, and the full backtrace of any exception.

#### ActiveRecord

```ruby
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'

  gem 'rails', '5.2.4.4'
  gem 'shrine', '3.3.0'
  gem 'sqlite3'
end

require 'active_record'
require 'shrine'
require 'shrine/storage/memory'
require 'down'

Shrine.storages = {
  cache: Shrine::Storage::Memory.new,
  store: Shrine::Storage::Memory.new
}

Shrine.plugin :activerecord

class MyUploader < Shrine
  plugin :derivatives
end

ActiveRecord::Base.establish_connection(adapter: 'sqlite3', database: ':memory:')
ActiveRecord::Base.connection.create_table(:posts) { |t| t.text :image_data }

class Post < ActiveRecord::Base
  include MyUploader::Attachment(:image)
end

post = Post.create(image: Down.download('http://theoldreader.com/kittens/600/400/'))

Marshal.dump(post)

# Traceback (most recent call last):
#	1: from reproduce_2.rb:36:in `<main>'
# reproduce.rb:36:in `dump': no _dump_data is defined for class Thread::Mutex (TypeError)


# HOWEVER!

Marshal.dump(post.class.find(post.id)) # works just fine

Marshal.dump(post.reload) # => no _dump_data is defined for class Thread::Mutex (TypeError)

Marshal.dump(post.image) # works just fine
```

[...]

PR #504, opened 2020-10-27
[#503] Fix Attacher marshalling with derivatives

Morozzzko:
Addresses #503 

This is a pretty naïve implementation to support Ruby's built-in `Marshal` behavior

This is a functional pull-request, however I have a couple of concerns about this implementation:

1. The tests don't really test anything. I wanted to check for equality, but attacher doesn't implement it. It's okay though.
2. The main problem is that this marshalling won't work well with plugins.

If any other plugins needs a complex setup in `Attacher#initialize`, we may break it after marshalling.

It also doesn't compose too easily.

So I'm thinking about an alternative solution – perhaps it would be better to explicitly _disable_ marshalling? This way, people won't have to spend as much time debugging as I did, and they'll know they have to use/write a custom plugin for that.

I'd love your feedback @janko. I can join discourse if we need to discuss it

### DIFF
--- a/lib/shrine/plugins/derivatives.rb
+++ b/lib/shrine/plugins/derivatives.rb
@@ -546,6 +546,31 @@ class Shrine
         def create_derivatives_on_promote?
           shrine_class.derivatives_options[:create_on_promote]
         end
+
+        # Support for proper marshalling
+        #
+        # It's impossible to marshal any models with this plugin
+        # as we can't marshal `Mutex` reliably.
+        def marshal_dump
+          instance_variables.reject do |name|
+            name == :@derivatives_mutex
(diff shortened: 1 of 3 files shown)


janko:
I like how this looks, thanks for working on it! 👍

I'll have to think about your concerns more, but it feels right to me that Shrine supports Marshalling out of the box.

janko:
After thinking about this over a longer period of time, it doesn't feel right that `derivatives` plugin is changing marshalling this way, as it doesn't give a clear API what other plugins should do if they add additional instance variables that hold non-marshallable values.

I would like to go with the alternative solution to allow disabling the mutex via a plugin option. I will push this shortly.

PR #507, opened 2020-11-14
derivation_endpoint typo fix

lordofthedanse:
* Fixes a typo (name of the plugin is derivation_endpoint, not derivation_endpoints)

### DIFF
--- a/doc/processing.md
+++ b/doc/processing.md
@@ -322,7 +322,7 @@ perform:
 ```rb
 # config/initializers/shrine.rb (Rails)
 # ...
-Shrine.plugin :derivation_endpoints, secret_key: "<YOUR_SECRET_KEY>"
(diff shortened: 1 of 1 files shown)


janko:
Thanks for the fix!

lordofthedanse:
@janko — you’re very welcome. I think (if I recall correctly) that there’s something like 2 places where this plugin is described (very cool, by the way), and it’s not super clear that you need the secret key in both the initializer and in the model (at least in one spot, maybe). Anyway, if you want — I’ll circle back to that and touch up the docs. Just let me know.

And I’ve been using Shrine for years, so a one character typo is probably pretty close to literally the least I could have done. Thanks for your work.

janko:
@lordofthedanse Happy to hear Shrine has been serving you well 😊 Yes, do send the PR, any documentation improvements are welcome 😉

Issue #508, opened 2020-11-16
undefined method `[]' for nil:NilClass 


dssjoblom:
I'm running into a strange issue, where it seems like options are not passed correctly to plugins (?). This seems to apply to almost all of them, e.g. url_options and store_dimensions.

### Expected behavior

That the code works normally on normal operations. 

### Actual behavior

This error can happen when doing almost anything, e.g. retrieving model.pic_url, where pic is the attachment. The message is undefined method `[]' for nil:NilClass when accessing options in a plugin. Here is another example:

```
begin
  rec.cover_pic = File.open( file ) if ok
rescue => e
  Utils.log_exception e, info: 'FAIL'
end
```

```
E, [2020-11-16T16:05:59.418508 #23534] ERROR -- : [acf054bb-3146-4eea-8c63-23e1ed1e47f7] undefined method `[]' for nil:NilClass
E, [2020-11-16T16:05:59.418681 #23534] ERROR -- : [acf054bb-3146-4eea-8c63-23e1ed1e47f7] /shared/vendor/bundle/ruby/2.7.0/gems/shrine-3.3.0/lib/shrine/plugins/store_dimensions.rb:75:in `extract_metadata'
/shared/vendor/bundle/ruby/2.7.0/gems/shrine-3.3.0/lib/shrine.rb:288:in `get_metadata'
/shared/vendor/bundle/ruby/2.7.0/gems/shrine-3.3.0/lib/shrine.rb:207:in `upload'
/shared/vendor/bundle/ruby/2.7.0/gems/shrine-3.3.0/lib/shrine.rb:108:in `upload'
/shared/vendor/bundle/ruby/2.7.0/gems/shrine-3.3.0/lib/shrine/attacher.rb:182:in `upload'
/shared/vendor/bundle/ruby/2.7.0/gems/shrine-3.3.0/lib/shrine/attacher.rb:117:in `attach'
/shared/vendor/bundle/ruby/2.7.0/gems/shrine-3.3.0/lib/shrine/plugins/validation.rb:41:in `attach'
/shared/vendor/bundle/ruby/2.7.0/gems/shrine-3.3.0/lib/shrine/attacher.rb:99:in `attach_cached'
/shared/vendor/bundle/ruby/2.7.0/gems/shrine-3.3.0/lib/shrine/plugins/validation.rb:34:in `attach_cached'
[...]

PR #509, opened 2020-11-18
Update documentation: plugin options being copied over

benkoshy:
Attempts to solve #508. I hope this helps.

Ben


janko:
Thanks for the contribution. I would prefer to describe this in the plugin system section of the getting started guide, I will push an update shortly.

PR #513, opened 2020-12-09
Change suggestion in migration in carrierwave.md

st-keller:
Using the :json or :jsonb datatype for the migration form CarrierWave may result in errors.
The json-structure will be read as a string and then the "to_hash"-method is missing.
Using :text works fine. 
The errors are difficult to debug - took me 3 days to figure this out.
Please look into it.

### DIFF
--- a/doc/carrierwave.md
+++ b/doc/carrierwave.md
@@ -275,8 +275,9 @@ Shrine. Let's assume we have a `Photo` model with the "image" attachment.
 First we need to create the `image_data` column for Shrine:
 
 ```rb
-add_column :photos, :image_data, :text # or :json or :jsonb if supported
+add_column :photos, :image_data, :text # please don't use :json or :jsonb here (even if supported) - this will result in errors during migration!
 ```
+
 
 ### 2. Dual write
 


hmistry:
Can you provide an example where you encountered errors when migrating the data from carrierwave text model to Shrine json? It will help us understand the edge cases and how to best solve it.

st-keller:
Unfortunately I didn't make screenshots or took notes - i just wanted to get rid of the error.

I'm on rails 6.0, my (local) postgres-db is on v13, i'm using the pg gem v1.2.3.
I have an existing model called "styleresource" with a "file"-attachment.
So i wrote a migration that added a "file_data"-column with the :json-datattype (thought it should be appropriate here - never used :json before).
The code in CarrierwaveShrineSynchronization ran just fine (no errors).
But as long as i had the :json-column i always got "undefined method `to_hash' for #<String:0x00...>"-errors when i tried to get the url of the file ("file.url" or "file_url").

Debugging pointed to this code in entity.rb as last call before the error occurred:
```
          # Returns the URL to the attached file.
          define_method :"#{name}_url" do |*args, **options|
            send(:"#{name}_attacher").url(*args, **options)
          end
```
So i assumed the value of the "file_data"-column was somehow read in as a string and not properly handled as a data-structure. Rolling back the migration that added a "file_data"-column with the :json-datattype and adding the same column as :text solved everything - what i took as proof for my assumption.

Hope this helps!
And thanks for all the work you put in the carrierwave.md - ist really great to get such wonderful assistance!

BKSpurgeon:
@hmistry Is it worth adding a more helpful warning if the wrong data type on the column is specified?

hmistry:
@st-keller Thanks that helps to understand the context. I was thinking maybe there was an edge case in carrier wave data stored that didn't convert properly to json format or something.

@BKSpurgeon No warning just yet. I'll need to test it out and see if where the issue is first.

janko:
@st-keller Thanks for the pull request. I'm sorry it took a long time to track the issue, and I appreciate the extra information, but I would prefer not to recommend against `:jsonb` type before we understand what was the issue.

If Shrine didn't detect the `:jsonb` data type correctly (which happens [here](https://github.com/lordofthedanse/shrine/blob/6a6cee2c99ca96837461d477d59c379c5f8669df/lib/shrine/plugins/activerecord.rb#L114-L117)), then it will assign a JSON string instead of a hash, and it will assume the column value is a string as well when it's parsing it, which should still work correctly. I think Active Record might have been reporting a different data type at the time of writing compared to at the time of reading.

In any case, until there is enough information to reproduce the issue, I'm closing this issue.

PR #514, opened 2020-12-16
Updates derivation_endpoint documentation

lordofthedanse:
* adds secret_key argument for derivation_endpoint plugin to docs
* renames and adds a note re: environment variable for clarity (i.e. you’re not using your Rails secret key)


janko:
Thanks for the contribution. I agree about the confusion the placeholder could cause, so I will accept that change.

One of the things I wanted to show in this example is that you can set default options such as `:secret_key` globally, so that you don't have to repeat it in uploader subclasses. So I will keep that part as it was.

PR #524, opened 2021-02-15
direct_s3.md: add JSON Bucket CORS config

FunkyloverOne:
It seems like they do not support XML format anymore in their new GUI


hmistry:
Thanks for the PR!

PR #525, opened 2021-02-22
Update getting_started.md

benkoshy:
Why this PR?

* Someone could copy/paste without thinking - leading to time spent debugging etc. hopefully this small line can obviate that.


janko:
Looks good, thanks for making these improvements 👍

PR #529, opened 2021-03-26
Fixes an argument in docs.

ordinathorreur:
Looks like this was a typo.


janko:
Thanks for the patch!

PR #531, opened 2021-04-05
Fixes incorrect class reference in the “Adding new derivatives” docs.

ordinathorreur:
The module seems to be incorrect and should be ImageProcessing instead.


janko:
Thanks for the patch 🙏

PR #542, opened 2021-06-12
fix: correct spelling

johngallagher:


janko:
Thanks for the patch!

PR #546, opened 2021-07-21
Stricter bracket expression for hex format test

aried3r:


janko:
Thanks, makes sense to check only valid hexadecimal numbers 👍🏻

PR #547, opened 2021-07-21
Add `infer_extension` convenience method to uploader

aried3r:
Same as `calculate_signature`. For a discussion (well, monologue) see [this discourse](https://discourse.shrinerb.com/t/how-to-call-infer-extension-in-custom-plugin-when-defined-as-a-dependency/502/2?u=aried3r).

Just a proposal, doesn't add anything but convenience and maybe more concise code. So if you don't think it should be exposed like this, feel free to just close this. :)


janko:
Looks good to me, thanks 👍🏻

PR #549, opened 2021-08-03
Add `signer` option to `derivation_endpoint` plugin

thibaudgg:
As discussed here: https://discourse.shrinerb.com/t/custom-derivation-endpoint-url-signer/495

### DIFF
--- a/lib/shrine/plugins/derivation_endpoint.rb
+++ b/lib/shrine/plugins/derivation_endpoint.rb
@@ -28,8 +28,8 @@ class Shrine
         uploader.opts[:derivation_endpoint] ||= { options: {}, derivations: {} }
         uploader.opts[:derivation_endpoint][:options].merge!(opts)
 
-        unless uploader.opts[:derivation_endpoint][:options][:secret_key]
-          fail Error, "must provide :secret_key option to derivation_endpoint plugin"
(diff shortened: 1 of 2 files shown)


thibaudgg:
@janko any update on merging my PR? 😇

janko:
I came back to this one now, sorry for the wait.

The API looks mostly good to me, I'm just wondering why the argument to the signer proc isn't the full URL (with query params), and instead the query parameters are passed separately? The query parameters need to be taken into account when signing, and `Aws::CloudFront::UrlSigner` seems to handle them just fine.

thibaudgg:
@janko if I remember correctly, I think it was because we needed to add the current time to the `expires_in` query parameters. We have the following custom signer logic:

``` ruby
  module CDN
    # Generate CloudFront signed URL
    def self.sign(url, expires_in:)
      signer = Aws::CloudFront::UrlSigner.new(
        key_pair_id: ENV['CLOUDFRONT_KEY_PAIR_ID'],
        private_key: ENV['CLOUDFRONT_PRIVATE_KEY'])
      signer.signed_url(url,
        expires: (Time.current + expires_in).to_i)
    end
  end
```

janko:
OK, finally got around to taking another look, and everything seems fine. At first I wasn't sure about the signer and dynamic option being differentiated by number of proc arguments, but now it seems fine to me. I'll merge it and add some documentation, then get version 3.5 released soon 👍🏻

PR #552, opened 2021-09-03
Correct usage of #atomic_persist when backgrounding

gogutan:
I added `current_file` just like https://github.com/shrinerb/shrine/commit/12b3e25b81000451e8429fceb7337501da60610b.


janko:
Looks good, thanks for the fix 😃

PR #564, opened 2022-01-21
Test with Ruby 3.1

aried3r:
- Fix CI errors
- Test with Ruby 3.1


aried3r:
```  
  1) Error:
plugin system::Shrine.plugin::when called globally#test_0001_allows the plugin to override base methods of core classes:
NoMethodError: undefined method `allocate' for Shrine::Attachment:Class

        assert_equal :foo, Shrine::Attachment.allocate.foo
                                             ^^^^^^^^^
    /home/runner/work/shrine/shrine/test/plugin_test.rb:46:in `block (4 levels) in <top (required)>'
```

I'm not quite sure why this fails and why only the last line.

https://github.com/shrinerb/shrine/blob/f62b7c3e7df4ab54f3a92fc009987d418645d10a/test/plugin_test.rb#L39-L46

I couldn't find anything immediately in either https://ruby-doc.org/core-3.1.0/NEWS_md.html or https://rubyreferences.github.io/rubychanges/3.1.html searching for `allocate`.

janko:
Thanks for the pull request!

Interesting 🤔 `Shrine::Attachment` is a subclass of `Module`, which in turn is an instance of `Class`, which should have `#allocate` instance method. I'll investigate.

janko:
I don't know why, but it seems that `Module.allocate` method has been removed in Ruby 3.1. I pushed a workaround for tests.

PR #565, opened 2022-01-27
validate_mime_type handle wildcard

randoum:
Allow using mime_types such as image/*

### DIFF
--- a/lib/shrine/plugins/validation_helpers.rb
+++ b/lib/shrine/plugins/validation_helpers.rb
@@ -171,7 +171,7 @@ class Shrine
         #     validate_mime_type_inclusion %w[audio/mp3 audio/flac]
         def validate_mime_type_inclusion(types, message: nil)
           validate_result(
-            types.include?(file.mime_type),
+            contains_mime_type?(types, file.mime_type),
             :mime_type_inclusion, message, types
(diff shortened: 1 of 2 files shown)


janko:
Thanks for the pull request, but I'm afraid this is not something I would like to add to Shrine. I consider accepting *all* image types as an anti-pattern, as it increases the attack surface for image processing, so I wouldn't like for Shrine to make it easy.

If you're using libvips, you could get the list of supported image types as follows:

```rb
Vips.get_suffixes
  .map { |ext| Marcel::MimeType.for(extension: ext) }
  .grep(/^image/)

#=> ["image/x-raw-panasonic", "image/x-portable-pixmap", "image/x-portable-graymap", "image/x-portable-bitmap", "image/jp2", "image/x-jp2-codestream", "image/png", "image/jpeg", "image/jpeg", "image/jpeg", "image/webp", "image/tiff", "image/tiff", "image/heic", "image/heif", "image/avif", "image/gif", "image/bmp"]
```

Alternatively, if you really wanted to allow all image types, you could do so with a custom validation:

```rb
Attacher.validate do
  unless file.mime_type.start_with?("image/")
    errors << "must be an image"
  end
end
```

PR #568, opened 2022-02-04
Update README.md example to avoid duplicate HTML element IDs

pond:
Resolves https://github.com/shrinerb/shrine/issues/567.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ allow users to upload files:
 
 ```erb
 <%= form_for @photo do |f| %>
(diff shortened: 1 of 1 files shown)


janko:
Thanks for the pull request. The `#field_id` method seems to have been added to Rails 7, and I would prefer examples to keep working with older versions of Rails. I would prefer a simpler alternative of just removing the `id` from the hidden field, I will push that shortly.

pond:
All good - I didn't know if Shrine had any use for a present ID in any other docs/recommendations or components - much simpler to remove & I see that the read-me is already updated with that approach. Thanks.

PR #570, opened 2022-02-13
Customize max multipart part

jpl:
Some s3 compatible cloud providers limit the number of multipart part during upload (example scaleway : https://www.scaleway.com/en/docs/storage/object/api-cli/multipart-uploads/).

This PR adds an options on the S3 driver to override the default max multipart part set to 10_000.


janko:
Looks good, thanks for the addition 👍🏻

PR #590, opened 2022-06-22
Fix: add bug report template

benkoshy:
### What is this?

A new workflow for github issues. People are required to carefully
choose questions and are directed to very specific templates and
formats.

image

If you choose a bug report: then instructions are given - try it out here: https://github.com/benkoshy/shrine/issues/new?assignees=&labels=bug%2Ctriage&template=1-bug-report.yml&title=%5BBug%5D%3A+

image

### Why?

* Far too many bug reports are produced without the appropriate code
reproduction examples. I have created / copied a new issues template
to make the instructions abundantly clear

* We want to encourage users to post questions on the github discussions
board, rather than opening issues on github.

With adequate reports, this will allow for contributors to help ease the maintenance burden. I hope this is helpful.


janko:
This looks great, thank you! ❤️

PR #599, opened 2022-10-05
Correct small typo in doc/plugins/derivatives.md

philippeapril:
The URL will be /fallbacks/ not /fallbacks.com/ (probably a bad find/replace/copy/paste of example.com


janko:
Thanks for the fix!

PR #600, opened 2022-10-12
Update domain in URLs of official Shrine articles

sliminas:
It seems like the old domain isn't used anymore or the articles were moved away from Github.

image


janko:
Thanks for the fixes! 👍🏻

PR #652, opened 2023-09-05
Make Shrine::DataFile truly top-level citizen

ixti:
`Shrine::DataFile` is already a top-level citizen, but available only of data uri plugin is loaded. For one use case I need to use DataFile directly when passing it to `assign` (creating attachment from the raw string), and this will allow do just that.

### DIFF
--- a/lib/shrine.rb
+++ b/lib/shrine.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require "shrine/data_file"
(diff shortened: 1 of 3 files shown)


janko:
Could you use `StringIO` directly for your use case?

ixti:
@janko yeah, after I've opened PR I just realized that I don't actually need DataIO or DataFile :D Sorry for the noize

PR #653, opened 2023-09-08
Duplicate the context hash when duplicating an Attacher

reidab:
We hit an issue following the upgrade to 3.5.0 where `context[:record]` in an `Attacher#finalize` call was set to an unpersisted duplicate of the record we were saving. This turned out to be because another part of our model save flow was duplicating the model for its own purposes and the `context` hash was shared between the original attacher and its duplicate.

The changes in fc6a3ed70aab9bff51ec1f0d269ad9553f4b8145  added a call to `attacher_copy.set_entity(self, name) if attacher_copy`. Because `@context` in both attachers point at the same object, this ends up with `context[:record]` for both records pointing to the new duplicate model.

### DIFF
--- a/lib/shrine/attacher.rb
+++ b/lib/shrine/attacher.rb
@@ -341,6 +341,13 @@ class Shrine
 
       private
 
+      # The copy constructor that's called on #dup and #clone
+      # We need to duplicate the context to prevent it from being shared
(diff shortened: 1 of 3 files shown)


janko:
Thanks for the patch, looks good 👍🏻

PR #660, opened 2023-11-03
Use lower case constants for headers (fixes Rack > 2 applications)

davidalejandroaguilar:
On Rack 3 https://github.com/rack/rack/issues/1592, it was decided to lowercase all response headers. Thus, all logic in the codebase that checks for headers coming from Rack > 2 applications is broken.

This PR:

* Uses lower case constants for headers.
* It does so by checking `Rack.release > "2"` so that we keep compatibility with Rack 2.x and 1.6.x.
  * This introduces `if` statements that look a bit ugly, but that was the pattern I found was being used within the codebase. 

# Note

I'm yet to write some specs for this but wanted to get this draft out for potential discussion.

### DIFF
--- a/doc/plugins/rack_response.md
+++ b/doc/plugins/rack_response.md
@@ -20,10 +20,10 @@ status, headers, body = uploaded_file.to_rack_response
 status  #=> 200
 headers #=>
 # {
-#   "Content-Length"      => "100",
-#   "Content-Type"        => "text/plain",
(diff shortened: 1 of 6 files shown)


tomasc:
see #682

janko:
Closing in favor of https://github.com/shrinerb/shrine/pull/682.

PR #689, opened 2024-06-14
forward options when calling #promote_cached to promote_block

4ndypanda:
If `Shrine::Attacher.promote_block { |**options| ... }` is declared with backgrounding plugin and we call `attacher.promote_cached(**options)`, the options were not being forwarded to the `.promote_block`.


janko (approved):
Makes sense, thanks!

PR #708, opened 2024-10-04
download_endpoint: Adds support for expiring URLs

davidwessman:
- Adds `expires_in` and `secret_key` to `download_endpoint` plugin
- This allows calling `attachment.download_url(expires_in: 5 * 60)`
  to generate a URL including `signature` and `expires_at` as query parameters. The timestamp is included in the signature and will only work before that timestamp.
- The URL is signed and verified with HMAC with SHA256 digest.


lib/shrine/plugins/download_endpoint.rb
```diff
@@ -65,14 +65,20 @@ def initialize(file)
...
         end
 
         protected
 
-        def path
-          file.urlsafe_dump(metadata: %w[filename size mime_type])
+        def path(expires_in: nil)
+          if expires_in || default_expires_in
+            reference = file.urlsafe_dump(metadata: %w[filename size mime_type])
+            expires_at = Time.now + (expires_in || default_expires_in)
```
davidwessman, line 77: Not 100% sure what to do here, I am used to ActiveSupports monkeypatch rails methods.
Maybe the argument should not be `expires_in` and instead be `expires_at` to avoid having to handle timezones and such.

lib/shrine/plugins/download_endpoint.rb
```diff
@@ -185,6 +205,11 @@ def get_uploaded_file(serialized)
     rescue Shrine::Error # storage not found
       not_found!
     rescue JSON::ParserError, ArgumentError => error # invalid serialized component
+      if @verifier_secret
```
davidwessman, line 208: I am not sure if we could add some special formatting / header to know that we should try to verify the data first.
Or should we always try to do that if the `@verifier_secret` is set?

test/plugin/download_endpoint_test.rb
```diff
@@ -25,10 +26,42 @@ def endpoint
...
+
+    assert_equal 200, response.status
+    assert_equal @uploaded_file.read, response.body
+    assert_equal @uploaded_file.size.to_s, response.headers["Content-Length"]
+    assert_equal @uploaded_file.mime_type, response.headers["Content-Type"]
+    assert_equal ContentDisposition.inline(@uploaded_file.original_filename), response.headers["Content-Disposition"]
+  end
+
+  it "does not return a file if expired" do
+    @uploader = uploader { plugin :download_endpoint, verifier_secret: SecureRandom.hex(64) }
```
davidwessman, line 56: Maybe I should also add a test where the URL is generated with one verifier_secret and then the request is processed by another?

adam12:
I suspect the use of AS::Verifier is going to be an issue. We might need to go with `OpenSSL::HMAC` directly.

davidwessman:
> I suspect the use of AS::Verifier is going to be an issue. We might need to go with `OpenSSL::HMAC` directly.

I'll try it out 🙂

lib/shrine/plugins/download_endpoint.rb
janko: The expiration timestamp should probably be in integer (aka UNIX) format in the URL, that way you don't have to worry about time zones.

davidwessman:
Now I rewrote it to use OpenSSL-directly 🙂 
Not sure if some of this functionality should be extracted to the `UrlSigner` class.

test/plugin/download_endpoint_test.rb
```diff
@@ -25,10 +25,42 @@ def endpoint
...
+    assert_equal @uploaded_file.mime_type, response.headers["Content-Type"]
     assert_equal ContentDisposition.inline(@uploaded_file.original_filename), response.headers["Content-Disposition"]
   end
 
+  it "raise error if using expires_in without any secret_key" do
+    io = fakeio("a" * 16*1024 + "b" * 16*1024 + "c" * 4*1024, content_type: "text/plain", filename: "content.txt")
+    @uploaded_file = @uploader.upload(io)
+    assert_raises Shrine::Error do
+      @uploaded_file.download_url(expires_in: 1000)
+    end
```
adam12, line 37: Did we want to assert on exception message here?

test/plugin/download_endpoint_test.rb
```diff
@@ -138,6 +170,34 @@ def endpoint
...
+  it "returns same download_url regardless of metadata order" do
+    @uploaded_file.data["metadata"] = { "filename" => "a", "mime_type" => "b", "size" => "c" }
+    url1 = @uploaded_file.download_url
+    @uploaded_file.data["metadata"] = { "mime_type" => "b", "size" => "c", "filename" => "a" }
+    url2 = @uploaded_file.download_url
+    assert_equal url1, url2
+  end
+
+  it "returns signature and expires_at when configured" do
+    secret = SecureRandom.hex(64)
```
adam12, line 182: I'd almost prefer a static secret here.

1. We don't have to generate it from entropy.
2. During test failures it's not going to cause issues recreating them.

Maybe something like `"A" * 64` assigned to a constant near the top of the test. WDYT?

adam12 (commented):
Nicely done adapting it for `OpenSSL::HMAC`. Few comments but overall looks good IMHO. Will leave final sign-off for Janko ofc.

davidwessman:
@janko could you allow the tests to run? They crash on my side because of `libmagic`

davidwessman:
@janko Sorry for a second (and last) ping, do you think there would be bandwidth for handling this PR within November?
Totally understandable if you do not know, trying to plan our alternatives with some deadlines coming up.
Thank you for great gems! 🙂

adam12:
Please rebase if willing. Then we can trigger tests to run and maybe get this over the finish line.

davidwessman:
> Please rebase if willing. Then we can trigger tests to run and maybe get this over the finish line.

Now I have rebased it, please trigger the tests 🙂

PR #721, opened 2025-04-16
Add Replace argument to refresh_metadata! in RefreshMetadata plugin

JacobGalati:
Currently `refresh_metadata!` merges the results of `Shrine.get_metadata` with the existing metadata. This means that when used in conjunction with the `add_metadata` plugin it doesn't handle changes that would remove unused or "deprecated" metadata.

For example, if the uploader previously looked like this

```rb

class ImageUploader < Shrine

  add_metadata :example_metadata do |io, context|
    'my custom metadata'
  end

  Attacher.derivatives do |original|
    {
      thumbnail:  ImageProcessing::Vips.source(original).resize_to_fill!(200, 200),
    }
  end
end
```
but was later updated to not add `example_metadata` to derivatives
```rb

class ImageUploader < Shrine

  add_metadata :example_metadata, skip_nil: true do |io, context|
    next nil if context[:derivative]

    'my custom metadata'
  end

  Attacher.derivatives do |original|
    {
      thumbnail:  ImageProcessing::Vips.source(original).resize_to_fill!(200, 200),
    }
  end
end
```
we end up with not changes after calling `refresh_metadata!` even though it would be reasonable to expect it to remove the `example_metadata`.

```rb
  uploaded_file(:thumbnail).metadata
  # {"filename"=>"example.jpg", "size"=>1024, "mime_type"=>"image/jpeg", "example_metadata"=>"my custom metadata"}
  uploaded_file(:thumbnail).refresh_metadata!(derivative: :thumbnail)
  uploaded_file(:thumbnail).metadata
  # {"filename"=>"example.jpg", "size"=>1024, "mime_type"=>"image/jpeg", "example_metadata"=>"my custom metadata"}
```

Now we can set `replace: true` to fully overwrite the existing metadata with the results of `Shrine.get_metadata`.

```rb
  uploaded_file(:thumbnail).metadata
  # {"filename"=>"example.jpg", "size"=>1024, "mime_type"=>"image/jpeg", "example_metadata"=>"my custom metadata"}
  uploaded_file(:thumbnail).refresh_metadata!(replace: true, derivative: :thumbnail)
  uploaded_file(:thumbnail).metadata
  # {"filename"=>"example.jpg", "size"=>1024, "mime_type"=>"image/jpeg"}
```

I would also be interested in looking at a way to handle automatically refreshing the metadata of any derivatives when `Shrine::Attacher.refresh_metadata!` is called, if the derivatives plugin is included. But I couldn't think of the best way to go about this.

### DIFF
--- a/doc/plugins/refresh_metadata.md
+++ b/doc/plugins/refresh_metadata.md
@@ -69,5 +69,25 @@ Any options passed in will be forwarded to metadata extraction:
 uploaded_file.refresh_metadata!(foo: "bar") # passes `{ foo: "bar" }` options to metadata extraction
 ```
 
+## Replacing Metadata
+
+By default the `#refresh_metadata!` method will merge the results into any existing metadata.
+
+```rb
+  uploaded_file.metadata["custom"] = "custom value"
+  uploaded_file.refresh_metadata!
+  uploaded_file.metadata
+  # returns {"filename"=>"example.jpg", "size"=>1024, "mime_type"=>"image/jpeg", "custom"=>"custom value"}
+```
+
+Passing `replace: true` will instead fully overwrite the existing metadata with the new metadata.
+
+```rb
+  uploaded_file.metadata["custom"] = "custom value"
+  uploaded_file.refresh_metadata!(replace: true)
+  uploaded_file.metadata
+  # returns {"filename"=>"example.jpg", "size"=>1024, "mime_type"=>"image/jpeg"}
+```
+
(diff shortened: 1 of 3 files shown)


adam12:
Looking at the source, we normally pass through all arguments to `refresh_metadata!` down the chain. Adding a `replace` argument means that we've now deviated from that API as the `replace` argument is specific to this method call.

I wonder if a new method would be more appropriate. `refresh_and_replace_metadata!` perhaps? Then we can keep the method signature as-is.

Maybe @janko can chime in if he has a moment.

Issue #725, opened 2025-06-10
Fix uploaded file URI::RFC3986_PARSER.make_regexp warning


bkuhlmann:
## Why

Hello. :wave: I'm getting the following warnings in my console:

``` 
/ruby/gems/3.4.0/gems/shrine-3.6.0/lib/shrine/uploaded_file.rb:53: warning: URI::RFC3986_PARSER.make_regexp is obsolete. Use URI::RFC2396_PARSER.make_regexp explicitly.
```

## How

This pops up when Ruby warnings are enabled (i.e. `$VERBOSE = true`). To fix, you'd only need to use `URI::RFC2396_PARSER.make_regexp` instead.

PR #729, opened 2025-06-17
Fixed URI default parser warnings

bkuhlmann:
## Overview

Necessary to resolve the following warnings showing up in output:

``` 
lib/shrine/uploaded_file.rb:53: warning: URI::RFC3986_PARSER.make_regexp is obsolete. Use URI::RFC2396_PARSER.make_regexp explicitly.
```

This can easily be replicated with `$VERBOSE = true` and/or with RSpec warnings enabled.

## Details

- Resolves Issue #725.


bkuhlmann:
OK, I've updated the code to be backwards compatible to as far back as Ruby 2.2.0. The commit message and above description has been updated accordingly.

PR #734, opened 2025-08-05
Document payload types for signature instrumentation

domcleal:
Adds the missing `algorithm` and `format` payload keys to the table and examples, also corrects the default log subscriber name.


adam12:
Thanks for the patch!

PR #740, opened 2025-10-10
Reorders put method upload selection to use @transfer_manager first

hms:
Reorders put method to use @transfer_manager if defined and then fallback to existing (but soon to be deprecated code)

I did my best to maintain the code formatting, especially in the tests. But it was rather challenging given my Rubymine configuration had very, very ideas about formatting than the project does.  I did my best to put things back by hand, but there were several area's where the changes were by spacing that did not show visually in Rubymine.

### DIFF
--- a/lib/shrine/storage/s3.rb
+++ b/lib/shrine/storage/s3.rb
@@ -239,25 +239,13 @@ class Shrine
       # for multipart uploads of large files.
       def put(io, id, **options)
         if @transfer_manager
-          transfer_manager_put(io, id, **options)
-        else
-          og_put(io, id, **options)
-        end
-      end
-
-      def transfer_manager_put(io, id, **options)
-        @transfer_manager.upload_stream(bucket: bucket.name, key: object_key(id), part_size: part_size(io), **options) do |write_stream|
-          IO.copy_stream(io, write_stream)
-        end
-      end
-
-      # Original (ganster) #put.
-      # Optimizes between object.put and object.upload_stream based on
-      # upload size.
-      def og_put(io, id, **options)
-        if io.respond_to?(:size) && io.size && io.size <= @multipart_threshold[:upload]
+          @transfer_manager.upload_stream(bucket: bucket.name, key: object_key(id), part_size: part_size(io), **options) do |write_stream|
+            IO.copy_stream(io, write_stream)
+          end
+        elsif io.respond_to?(:size) && io.size && io.size <= @multipart_threshold[:upload]
           object(id).put(body: io, **options)
-        else # multipart upload
+        else
+          # multipart upload old API
           object(id).upload_stream(part_size: part_size(io), **options) do |write_stream|
             IO.copy_stream(io, write_stream)
           end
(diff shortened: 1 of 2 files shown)


hms:
@adam12 

Unfortunately fair.  But to be honest, it would be helpful to have at least something to reference so I can configure my editor to honor this projects defaults.

I guess I'll reimplement the PR using VIM to avoid unexpected reformatting.

jrochkind:
> If the download functionality was also migrated to the transfer_manager, would Shrine still need the Down Gem?

Yes, among other reasons because Shrine uses more than just S3 storage adapters. 

But I too have been suspecting some memory issues in Down, although they may be just kind of unavoidable in ruby is also what I was maybe arriving at (the memory issues may not be particular to Down, but endemic to any attempt to copy large files in ruby). But I'm very curious as to what you may find, although it's not about this issue/PR, maybe open up a Discussion?

hms:
@jrochkind Thanks for the gentle reminder that sometimes the problem is bigger than the one impacting me.... I really wasn't thinking outside of S3 at all.

When I find the time to dig into Down again, I'll open up a ticket over there and reference it here.

hms:
@janko @adam12 

Are you folks waiting on more changes from me to move this PR forward? 

hms

janko:
@hms I realized I don't actually understand what this PR improved/fixed. Now that everything goes though transfer manager, multipart upload is being used even for small files. Why is that better than using single `#put` request (which was never deprecated) for small files?

hms:
@janko 

My understanding of the AWS S3 TransferManager class is that it handles small and large uploads, doing the right thing for performance and reliability.

It's been a while since this patch request and I don't have the pages that I reviewed to reference for this discussion, but I vaguely remember reading something that suggested that the TransferManager was going to be the preferred path and the other API's were going to be deprecated over time.  I may have just hallucinated that memory or simply got it wrong and misunderstood, but unless it's significantly slower in important use cases, my rational is there is safety in using the SDK recommended path.

If it is screwing up a common use-case, my apologies and I'm happy to supply a revert / reordering of the branches.


hms

janko:
From the [docs](https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/S3/TransferManager.html#), it seems that `Aws::S3::TransferManager` uses multipart uploads unconditionally. This isn't efficient for smaller files IMO, because you're paying for request overhead 3 times instead of once, and it breaks the `:multipart_threshold` storage option assumption (it doesn't have effect currently).

I'm revisiting this because I've seen some multipart errors in Sentry for our app at work after upgrading to latest version of Shrine, and I don't really understand why.

Given that it seems the original assumption was that `Aws::S3::TransferManager` will retain the previous behavior, I'll revert to keep using `#put` for smaller files, no need to send a new PR. It seems that Active Storage does exactly the same, which is a strong indicator that it's the right implementation.

moofkit:
@janko thanks for handling this. It's actually more just a performance penalties.
Here is our cost reports on one of the bucket where average file size is around 200kb:

shrine3_7_0

It increased spendings 2x.

Thanks for the fix

janko:
Just released version 3.8.0 with the revert.

Issue #723, opened 2025-05-06
Feature: Keep derivatives when attaching a new file


fnordfish:
Background:

I'm working on a project where files and their derivatives can be updated independently.  
That means, that updating a "source" file doesn't necessarily mean, that its derivatives should be cleared/re-generated.

Here's my monkey-patch that allows me to do that in my project:

```ruby
class Shrine
  module Plugins
    # Keep derivatives when attaching a new file to a record
    # and the :keep_derivatives plugin option is set to true.
    module Derivatives
      module AttacherMethods
        # Clears derivatives when attachment changes.
        #
        #     attacher.derivatives #=> { thumb: #<Shrine::UploadedFile> }
        #
        #     # Original
        #     attacher.change(file)
        #     attacher.derivatives #=> {}
        #
        #     # With keep_derivatives: true
        #     attacher.change(file)
        #     attacher.derivatives #=> { thumb: #<Shrine::UploadedFile> }
        def change(*)
          result = super
          set_derivatives({}) unless shrine_class.derivatives_options[:keep_derivatives]
          result
        end
      end
    end
  end
end
```

janko:
@fnordfish Sounds good, would you mind submitting a PR for this?

janko:
Merged https://github.com/shrinerb/shrine/pull/743.

PR #743, opened 2025-12-05
Add keep_derivatives option to derivatives plugin to keep derivatives when attaching a new file

fnordfish:
see #723

There is only one minimal test :/


janko:
Looks good, thanks!

PR #749, opened 2026-05-28
Forward promote options to derivatives

### LINKED ISSUE #745: Attacher#promote does not forward **options to promote_derivatives in derivatives plugin
## Description

In `lib/shrine/plugins/derivatives.rb`, `Attacher#promote` accepts `**options` but does not pass them through to `promote_derivatives`:

```ruby
# lib/shrine/plugins/derivatives.rb L162-166 (v3.6.0)
def promote(**options)
  super
  promote_derivatives          # <-- **options not forwarded
  create_derivatives if create_derivatives_on_promote?
end
```

`promote_derivatives` already accepts `**options` and forwards them to `upload_derivative`, so the fix should be:

```ruby
def promote(**options)
  super
  promote_derivatives(**options)
  create_derivatives if create_derivatives_on_promote?
end
```

## Impact

Upload options (e.g., `upload_options` for setting `Cache-Control` headers or other storage-level metadata) are applied to the main file during promotion but silently lost for derivatives. This means derivatives promoted from cache to permanent storage do not receive the intended storage options.

## Steps to reproduce

```ruby
Shrine.plugin :derivatives
Shrine.plugin :upload_options, store: -> (io, **) {
  { cache_control: "public, max-age=31536000, immutable" }
}

attacher = Shrine::Attacher.new

# Observe that upload_options are applied when promoting the main file (via super),
# but NOT when promoting derivatives, because **options is dropped.
attacher.promote
```

## Versions

- Shrine 3.6.0
- Ruby 3.4

milkcocoa:
## Summary

- Forward `Attacher#promote` keyword options to `promote_derivatives` in the derivatives plugin.
- Add a regression test covering upload options applied to both the main file and cached derivatives during promotion.

## Why

`promote_derivatives` already accepts and forwards upload options, but `promote` dropped them when calling it. This caused storage-level upload options to apply to the main file while being lost for promoted derivatives.

Fixes #745.

## Test

- `CI=1 BUNDLE_PATH=/private/tmp/shrine-bundle bundle exec rake test`

### DIFF
--- a/lib/shrine/plugins/derivatives.rb
+++ b/lib/shrine/plugins/derivatives.rb
@@ -161,7 +161,7 @@ class Shrine
         #     attacher.stored?(attacher.derivatives[:thumb]) #=> true
         def promote(**options)
           super
-          promote_derivatives
+          promote_derivatives(**options)
           create_derivatives if create_derivatives_on_promote?
         end
 
--- a/test/plugin/derivatives_test.rb
+++ b/test/plugin/derivatives_test.rb
@@ -327,6 +327,17 @@ describe Shrine::Plugins::Derivatives do
         assert_equal "foo", @attacher.file.id
       end
 
+      it "forwards promote options to derivatives" do
+        @shrine.plugin :upload_options, store: -> (_io, options) { { foo: options[:foo] } }
+
+        @attacher.attach_cached(fakeio)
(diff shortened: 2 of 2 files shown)


janko (approved):
Looks good, thanks!

PR #751, opened 2026-07-24
Add :max_size option to data_uri plugin

jalrax:
The data_uri plugin currently has no way to limit the size of the uploaded content — the whole payload gets decoded into memory before validations get a chance to run. remote_url has `:max_size` for exactly this, so this PR adds the same option to data_uri.

I'm aware the general recommendation is to cap request body size at the web server, and we do that too. But `client_max_body_size` is one global knob — in our app different uploaders have different limits (20 MB for audio, 30 MB for images), and the only place to express a per-uploader cap is Shrine itself. Without it, a payload that's over the uploader's limit but under the global one still gets fully decoded just to be thrown away by `validate_max_size`.

Following remote_url, `:max_size` is the maximum size of the decoded content. The check runs inside `parse_data_uri`, before `post_match` copies the payload out of the URI — on a 30 MB data URI that copy alone was 30 MB of RSS. The size is computed on the URI itself: exact for base64 (counting alphabet characters, so line-wrapped payloads work too), and the encoded length for percent-encoded content, which decoding can only shrink. Rejecting a 30 MB URI now costs 0 MB and 17 ms.

Oversized content fails with `ParseError`, so the `assign_data_uri` error contract is unchanged. Since `ParseError` now has two causes, `:error_message` also receives the error, the same way remote_url does it.

We've been running this in production as a monkey patch; upstreaming it in case it's useful to others.

### DIFF
--- a/doc/plugins/data_uri.md
+++ b/doc/plugins/data_uri.md
@@ -37,18 +37,22 @@ attachment column. You can change the default error message:
 ```rb
 plugin :data_uri, error_message: "data URI was invalid"
 plugin :data_uri, error_message: ->(uri) { I18n.t("errors.data_uri_invalid") }
+plugin :data_uri, error_message: ->(uri, error) { I18n.t("errors.data_uri.#{error.message}") }
 ```
 
 ## Maximum size
 
-The `:max_size` option sets the maximum allowed size of the decoded content.
-The size is determined from the encoded payload, so oversized content is
-rejected before it gets decoded into memory. Parsing fails the same way as
-for a malformed URI:
+It's a good practice to limit the maximum size of the data URI content:
 
 ```rb
 plugin :data_uri, max_size: 10*1024*1024 # 10 MB
 ```
+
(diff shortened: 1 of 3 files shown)


janko:
@jalrax Thanks for the patch! Do you think it would be better avoid scanning/parsing entirely if the data URI is longer than `:max_size`?

jalrax:
@janko 

Good catch — post_match copies the payload out, so a 30 MB data URI cost 30 MB of RSS before the check even ran.

Comparing uri.bytesize directly wouldn't work: `:max_size` is the decoded size (like remote_url), and base64 inflates by 4/3, so a valid 30 MB file arrives as a ~40 MB URI. The check now runs inside `parse_data_uri` before post_match, computing the size on the URI itself — exact for base64, encoded length for percent-encoded. Same 30 MB URI: 0 MB, 17 ms.

Also, ParseError has two causes now, so ` :error_message` receives the error as well, like in remote_url.

janko (approved):
Looks good, thanks! 👍🏻

